PR #20, opened 2021-03-15
Replace rem with em

barryceelen:
Using rem-based sizing does not / may not play well in the context of the block editor, the base font size in the admin might be different making it harder to style blocks in the editor identically to the front end.

### DIFF
--- a/themes/10up-theme/assets/css/frontend/base/wordpress.css
+++ b/themes/10up-theme/assets/css/frontend/base/wordpress.css
@@ -2,13 +2,13 @@
 
 .alignleft {
 	float: left;
-	margin-right: 1rem;
+	margin-right: 1em;
 	text-align: left;
 }
 
 .alignright {
 	float: right;
-	margin-left: 1rem;
+	margin-left: 1em;
 	text-align: right;
 }
 


joesnellpdx:
@barryceelen Thanks for the PR. I'll have to investigate this as the best solution. REMs definitely don't play nice in the admin without internally connecting the admin root font size to the front end. EMS can have additional cascading issues if component font sizes are changed, then nested within eachother.

smy315:
@barryceelen this file (css/frontend/base/wordpress.css) is only for the frontend so it doesn't impact the WordPress admin.

barryceelen:
@smy315 the issue I'm trying to address here is really that when the front end styles are included in the editor (as the css/frontend/base/wordpress.css might be), any rem-based styles do not work well in the admin. I'm seeing this happen on a client site where the theme's custom block rem-based font sizes are not the same in the block editor as they are on the front end due to the base font size on the body not being identical on the front and back end.

smy315:
@barryceelen i understand the rem issue but CSS code within wordpress.css should never be included in the admin. WordPress already has styles for .alignleft and .alignright via Core. The point of this wordpress.css file is to provide CSS for the frontend only. Not everything we add on frontend should be included within admin and this is a good example.

I actually worked on this file getting merged in few years ago: https://github.com/10up/plugin-scaffold/pull/61. Main point of this was to provide support on the frontend that WordPress admin already had.

barryceelen:
Closing this out as there is nothing inherently wrong with using rem. Styling discrepancies only occur when the theme defines the root font size on the html element as anything else than the admin does, for example by setting it to 62.5% to simplify using rem values. The theme scaffold in this repo does no such thing. Thanks all!

PR #22, opened 2021-04-01
Disable support options as a default starting point.

ryanwelcher:
### Description of the Change

This PR disables all `supports` non-standard options as a starting point for a new block. Also implicitly sets items that are true by default. 

### Benefits
By ensuring that using a feature requires a step, the block being created will be more purpose-built and consistent. By displaying all options ( even ones that are set by default ) will also show what is possible for the block.

### Possible Drawbacks
Items such as custom class names that are usually enabled by default will require a step to enable.

### Checklist:


tlovett1:
Great idea

PR #23, opened 2021-04-07
Feature/lerna

nicholasio:
### Description of the Change

This PR introduces lerna https://github.com/lerna/lerna for handling npm packages and scripts execution from the root folder.

### Alternate Designs

Not use lerna and run npm install individually on each theme/plugin.

### Benefits

- Hoist all deps to the root folder, avoid duplication, saving storage and reduce the time it takes to install deps.
- Facilitates running scripts across all themes/plugins. Running `npm run build` at the root will run build in all themes/plugins in parallel.

### Possible Drawbacks

Hoisting deps to the root could introduce issues in setups that are not conformant to the node module resolution algorithm but it's unlikely to affect us. Additionally, if this solution doesn't work for some reason for a project, there's probably something that is being done wrong.

Worst-case scenarios projects can opt-out for hoisting deps.

### Checklist:


README.md
```diff
@@ -9,6 +9,20 @@ It contains a bare bones theme and must use plugin for you to base your developm
...
 3. Compiling, minifying, bundling, etc. of JavaScript and CSS is all done by [10up Scripts](https://github.com/10up/10up-scripts). 10up Scripts is included as a dev dependency in both the plugin and theme. If you want to develop on the theme (and vice-versa the plugin), you would navigate to the theme directory in your console and run `npm run start` (after running `npm install` first of course). Inside `package.json` edit `@10up/scripts.devURL` to your local development URL for BrowserSync to work properly. `@10up/scripts.entry` are the paths to CSS/JS files that need to be bundled. Edit these as needed.
+4. [lerna](https://github.com/lerna/lerna) is used to managed npm dependencies. The benefits of using lerna is that we can hoist all dependencies to the root folder and avoid installing duplicate dependencies, saving time and space. By default `lerna` is aware of all themes and `mu-plugins/10up-plugin`, if you are building a new plugin/theme make sure that `lerna` is aware of it by adding it to the `lerna.json` file. See the example below:
```
bengreeley:
```suggestion
4. [lerna](https://github.com/lerna/lerna) is used to manage npm dependencies. The main benefit of using lerna is that we can hoist all dependencies to the root folder and avoid installing duplicate dependencies, saving time and space. By default `lerna` is aware of all themes and `mu-plugins/10up-plugin`, if you are building a new plugin/theme make sure that `lerna` is aware of it by adding it to the `lerna.json` file. See the example below:
```

bengreeley:
@nicholasio I really like this addition, I think it makes a lot of sense and will help prevent the need for projects to come up with their own convoluted setup instructions docs.

tlovett1:
Tested and it works for me.

Issue #21, opened 2021-03-17
Set up lint-staged + hustky


nicholasio:
The wp-scaffold is missing the husky + lint-staged setup which runs eslint and phpcs on every commit.,

I suggest adding that to the package.json at the root so that we have one set up linting both plugin and theme.

PR #25, opened 2021-04-08
Feature/add husky lint staged

darylldoyle:
### Description of the Change

Adds lint-staged and husky to run on the pre-commit hook.

### Alternate Designs

N/A

### Benefits

Makes sure that files are linted prior to being committeded

### Possible Drawbacks

More config to maintain

### Verification Process

Tested on local machine

### Checklist:

### Applicable Issues

#21 

### Changelog Entry

- Added lint-staged and husky


.lintstagedrc.json
```diff
@@ -0,0 +1,14 @@
+{
```
nicholasio, line 1: Any reason not to store the lintsage config in package.json that's how we used to have this set up in the past.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
nicholasio, line 1: Any benefit of adding the pre-commit hook here instead of `package.json`? Also not sure what line 2 is doing here.

.lintstagedrc.json
```diff
@@ -0,0 +1,14 @@
...
+    "*.css": [
+      "10up-scripts lint-style"
```
nicholasio: can you update this to `10up-toolkit` as per https://github.com/10up/wp-scaffold/pull/35

package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://project-git-repo.tld"
   },
   "scripts": {
-    "prepare": "lerna bootstrap --hoist",
+    "prepare": "lerna bootstrap --hoist && husky install",
```
nicholasio, line 11: Is it really necessary to manually call `husky install`? I thought husky would automatically install himself.

darylldoyle: Husky dropped their auto-install. They now recommend that you use the install command on prepare: https://blog.typicode.com/husky-git-hooks-autoinstall/

.lintstagedrc.json
darylldoyle: Just personal preference of being able to see where different configs live, I'm happy to move it into the package.json if you'd prefer?

.husky/pre-commit
darylldoyle: This is where husky stores the hooks now: https://typicode.github.io/husky/#/?id=create-a-hook

Line two exists in all these new style husky hooks and calls the husky shell script which is installed at `.husky/_/husky.sh` when `husky install` is run.

Here's an example of the same in the husky repo: https://github.com/typicode/husky/blob/main/.husky/pre-commit

darylldoyle: There's more info on how this changed from v4-v6 available here: https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v6

.lintstagedrc.json
nicholasio: I think that's fine, let's keep it this way since we're setting up husky hooks outside of package.json as well.

nicholasio (approved):

Issue #27, opened 2021-04-08
Add jest unit tests for example block 


Shelob9:
By default, there are no Jest tests for the example-block. To encourage testing, I want to add some basic tests for the React components that are used to make the block.  I am mainly adding snapshot tests, as those will be useful immediately on the project and will be easy to update.

I am working on a PR, but ran into #26

Issue #26, opened 2021-04-08
Babel error when running Jest tests


Shelob9:
**Describe the bug**

When using the `npm run test` command, from the root of the repo or the root of the theme, I get errors like this and the tests fail:

```
[BABEL] /home/josh/wp-scaffold/themes/10up-theme/includes/blocks/example-block/test.test.js: Unknown option: .wordpress. Check out https://babeljs.io/docs/en/babel-core/#options for more information about options.
    - Maybe you meant to use
    "preset": [
      ["@10up/babel-preset-default", {
      "wordpress": true
    }]
    ]
    To be a valid preset, its name and options should be wrapped in a pair of brackets

      at throwUnknownError (node_modules/@babel/core/lib/config/validation/options.js:131:27)
      at node_modules/@babel/core/lib/config/validation/options.js:116:5
          at Array.forEach (<anonymous>)
      at validateNested (node_modules/@babel/core/lib/config/validation/options.js:92:21)
      at validate (node_modules/@babel/core/lib/config/validation/options.js:83:10)
      at node_modules/@babel/core/lib/config/full.js:320:36
      at cachedFunction (node_modules/@babel/core/lib/config/caching.js:52:27)
          at cachedFunction.next (<anonymous>)
      at evaluateSync (node_modules/gensync/index.js:251:28)
      at sync (node_modules/gensync/index.js:89:14)

```
**Steps to Reproduce**
1. Download wp-scaffold
2. Add "themes/10up-theme/includes/blocks/example-block/test.test.js"
3. Add any test to that dir `test( 'something', ()=> { expect(1).toBe(1)`;
5. Run `npm run test` from the theme's directory.
6. Add a test to the mu-plugin
7.  See two of the same error

**Environment information**
 - Device: Window Laptop
 - OS: Windows 10 Pro with WSL/2. Issue is happening with  Ubuntu 20.04.2 LTS
 - Tested with Node 12 and Node 14.
 - NPM 6.14.11

PR #15, opened 2021-02-24
Setting up test suite and adding an example test

nicholasio:
### Description of the Change

This sets up jest tests with 10up-scripts and adds an example test of the example block.

### Benefits

Unit tests infrastructure and a simple example of a jest test.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

PR #28, opened 2021-04-08
Add example tests for blocks

Shelob9:
### Description of the Change

To encourage testing React components, this PR adds basic tests for example block. There are two tests:

* A snapshot test.
* A test that a changing the rich text input calls `setAttributes`

These are good first tests and show some concepts.

### Alternate Designs

I installed @testing-library/react in the theme. There are alternatives for example Enzyme that I could have used.

### Benefits

Blocks start with useable tests
### Possible Drawbacks

Not everyone is familiar with React testing. We have to keep @testing-library up to date.

### Verification Process

I need to fix/ workaround #26 before testing this.

- How did you verify that all new functionality works as expected?
- How did you verify that all changed functionality works as expected?
- How did you verify that the change has not introduced any regressions?

### Checklist:

- [ ] My code follows the code style of this project.

### Applicable Issues

#27 

### Changelog Entry

ADDED: Example Jest unit tests in example block.

### DIFF
--- a/themes/10up-theme/includes/blocks/example-block/block.test.js
+++ b/themes/10up-theme/includes/blocks/example-block/block.test.js
@@ -1,0 +1,21 @@
+import Edit from './edit';
+
+import {render,act,fireEvent} from '@testing-library/react';
+describe( 'Basic block', () => {
+    test( 'Edit component matches snapshot ', () => {
+        const setAttributes = jest.fn();
+        const {container} = render(<Edit setAttributes={setAttributes} className={'one two'} attributes={{customTitle: 'Title'}} />)
+        expect(container).toMatchSnapshot();
+    });
+
+    test( 'Edit component calls setAttibutes ', () => {
+        const setAttributes = jest.fn();
+        const customTitle = 'Title';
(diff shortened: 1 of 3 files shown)


fabiankaegy:
Hey @Shelob9, this is very cool. How did you get seeping the fact that most of the components that get used in the Blocks Edit React component are actually externals that jest won't have access to?

Since we want to avoid having to include our own version of all the @wordpress packages in our themes and plugins. 

That's always what I find most challenging.

Shelob9:
@fabiankaegy Thanks. 

> How did you get seeping the fact that most of the components that get used in the Blocks Edit React component are actually externals that jest won't have access to?

Good question. So, I havn't gotten there yet, got stuck trying to figure out #26  . In the past, I've made the WordPress dependency extraction webpack plugin conditional, so it doesn't run during testing. I've also imported the scripts in jest's setup script and used that to set window.wp.

nicholasio:
I did a similar PR a while ago https://github.com/10up/wp-scaffold/pull/15/files

@Shelob9 The fact that pretty much all components are externals is the reason I wanted to build a `jest-gutenberg` package that would render the block-editor in jsdom so that we don't have to mock things cc @fabiankaegy 

While I like the idea of adding example tests to the scaffold I think that in practice, engineers will get stuck if they have to mock everything.

PR #31, opened 2021-04-30
Add allowed blocks filter / array to scaffold

### LINKED ISSUE #30: Add Allowed Blocks to ensure only intentional blocks are enabled
**Is your enhancement related to a problem? Please describe.**
- Ensure only intended blocks are enabled from a commonly used list of blocks
- Ensure this pattern if followed on projects going forward to prevent *new* core blocks from being enabled without passing engineering (styling, functionality) 

Allowed blocks list enabled

joesnellpdx:
### Description of the Change

Add allowed blocks array to project

- Ensure only intended blocks are enabled from a commonly used list of blocks
- Ensure this pattern if followed on projects going forward to prevent new core blocks from being enabled without passing engineering (styling, functionality)

### Verification Process

Ensured all blocks listed in the array function as expected.

### Checklist:

### Changelog Entry

ADDED: Allowed blocks array


themes/10up-theme/includes/blocks.php
```diff
@@ -22,6 +22,8 @@ function setup() {
 
 	add_action( 'enqueue_block_editor_assets', $n( 'blocks_editor_styles' ) );
 
+	add_filter( 'allowed_block_types', $n( 'allowed_block_types', 10, 2 ) );
```
darylldoyle: @joesnellpdx the closing bracket is in the wrong place here, it should be:

```php
add_filter( 'allowed_block_types', $n( 'allowed_block_types' ), 10, 2 );
```

darylldoyle:
@joesnellpdx adding this filter as-is will remove all custom blocks from a site. I'm using this on a site in dev at the moment. We can fix this issue via a change to the `register.php` within a block directory:

```php
/**
 * Register the block
 */
function register() {
	$n = function( $function ) {
		return __NAMESPACE__ . "\\$function";
	};
	// Register the block.
	$result = register_block_type_from_metadata(
		DEVRANET_THEME_BLOCK_DIR . '/example-block', // this is the directory where the block.json is found.
		[
			'render_callback' => $n( 'render_block_callback' ),
		]
	);

	// If we got a successful registration, filter the allowed blocks.
	if ( $result instanceof \WP_Block_type ) {
		add_filter(
			'allowed_block_types',
			function( $allowed_block_types, $post ) use ( $n, $result ) {
				// If the allow_block function returns true, add the block to the allow list.
				if ( $n( 'allow_block' )( $post ) ) {
					$allowed_block_types[] = $result->name;
				}

				return $allowed_block_types;
			},
			20,
			2
		);
	}
}
```

This uses a new function called `allow_block()` that receives the current post object. The `allow_block()` function should return a boolean. If `true` the block will be added to the allow list, if false it won't be.

We run the `allowed_block_types` filter with a priority of `20`, so that it runs after the core block restrictions.

```php
/**
 * Whether or not this block should be available to select.
 *
 * @param \WP_Post $post The current post object.
 *
 * @return bool
 */
function allow_block( $post ) {
	return true;
}
```

This is the `allow_block()` function. It receives `$post` so that more advanced logic can be used, rather than just always adding the block. This will allow the engineer to enable the block for certain posts/post types/any other logic that can be deducted from a post object.

Hopefully that's of some use!

themes/10up-theme/includes/blocks.php
```diff
@@ -90,6 +92,45 @@ function blocks_editor_styles() {
...
+ * Allowed blocks in editor
+ * See reference of available blocks here: https://wordpress.org/support/article/blocks/
+ *
+ * @param array $allowed_block_types core array
+ * @return array  $allowed_blocks modified array under right conditions.
+ */
+function allowed_block_types( $allowed_block_types ) {
+
+	$allowed_blocks = [
+		'core/paragraph',
```
smy315: @joesnellpdx where is this list from? For example, I see that core/embed is missing and YouTube embeds are one of the most commonly used blocks. I def think this allowed block filter is useful but should make sure we don't blacklist commonly used blocks out of the box.

Also, include a comment or maybe even add the 10up-plugin (mu-plugins/10up-plugin) commented out to this list as an example of including custom blocks? Custom blocks will not work if we add this filter so we should provide some sort of documentation or direction.

themes/10up-theme/includes/blocks.php
```diff
@@ -90,6 +92,45 @@ function blocks_editor_styles() {
...
+		'core/image',
```
smy315: Also, alpha sort this list to easily check what is on this list? Took me a bit to look through the list to see if anything important like core/embed was missing.

smy315 (changes requested):

themes/10up-theme/includes/blocks.php
joesnellpdx: @smy315  List came from UX team and some others. You are right, embed should be in there - added that back. 

Great advice to also add better guidance for block library blocks and custom blocks - this has been updated.

joesnellpdx:
@darylldoyle 

GREAT feedback / tips! I made updates per your suggestions. I'd love you to give it a glance if you have time! Thanks!

ryanwelcher:
@joesnellpdx I wonder how often we're building blocks for a project and then limiting them? To be clear, this is a cool feature but just wondering if it belongs as part of the default for the scaffold. Perhaps it's better as a snippet somewhere that can be referenced?

joesnellpdx:
> @joesnellpdx I wonder how often we're building blocks for a project and then limiting them? To be clear, this is a cool feature but just wondering if it belongs as part of the default for the scaffold. Perhaps it's better as a snippet somewhere that can be referenced?

@ryanwelcher The use case here is making sure we are limiting the core blocks to start intentionally - so we can avoid "new core blocks" that are unstyled/untested getting into the editorial environment. 

Per the custom blocks, the additions are intended to get them into the allowed list (same as the selected block library plugin items). 

Does that make sense? Let me know if a quick call would be a better realm to discuss.

CC: @fabiankaegy - would love to get your eyes here too.

tlovett1:
Are we intentionally leaving out the `core/html` block?

joesnellpdx:
> Are we intentionally leaving out the `core/html` block?

@tlovett1 I did, but I'm not overly opinionated about it. If you feel it should be included by default, I'll add it. 

My thoughts: do most clients wan the ability for editors to add random html to pages/posts that could effect layout, design and potentially cause issues (layout, security, etc)? I'd think they don't, but maybe that is making too strong of an assumption.

ryanwelcher:
> @ryanwelcher The use case here is making sure we are limiting the core blocks to start intentionally - so we can avoid "new core blocks" that are unstyled/untested getting into the editorial environment.
> 
> Per the custom blocks, the additions are intended to get them into the allowed list (same as the selected block library plugin items).
> 

@joesnellpdx I don't disagree with this approach, I would caution that that code will not allow any blocks introduced by third-party plugins ( i.e Jetpack ) without a code change.

Perhaps the approach should focus on limiting only core blocks with the assumption that non-core blocks are intended to be in place because they were either a) created by the project team or b) provided by a plugin that has been purposefully chosen to provide them. Taking this approach would remove the need to add the registered custom blocks to the array it will also remove the need to add blocks provided by managed hosts such as VIP.

tlovett1:
Yea I think I agree with @ryanwelcher. Playing around with the scaffold the first thing that tripped me up is I couldn't figure out why the custom HTML block wasn't showing up. The second thing was that I need to allow my custom block even though I am registering it. I think we should do the following:
1. As Ryan said, allow all plugin blocks by default as it could be extremely confusing to engineers trying to figure out why a plugin block isn't showing up.
2. Custom blocks created in the theme should be allowed automatically. I shouldn't have to manually allow them.
3. Extremely obvious documentation around what blocks are allowed by default and where to change that.

joesnellpdx:
@ryanwelcher @tlovett1 

I agree. That is what I was trying to solve with those extra filters in `register.php` - to allow custom blocks out of the box (as it is now any custom block added to the theme - if done like the example block - will be automatically allowed). This, however, doesn't account for blocks added by plugins - so I definitely understand that issue. 

Also, for clarity, @tlovett1 when you say "custom html block" you are referring to the [core "custom html block"](https://wordpress.org/support/article/custom-html/) - no need to respond but the word "custom" threw me a second. 

Ideally, I want that "allowed_blocks" array ONLY pertain to core blocks - and everything else is automatically allowed (that was my intention and I think I was confused by the comments). The issue is that it effects all blocks out of the box (from what I can tell).

I'll investigate further on an alternative solution so it so that only effects core blocks. 

TODO:

1. Allow all plugin blocks by default as it could be extremely confusing to engineers trying to figure out why a plugin block isn't showing up.
2. Custom blocks created in the theme should be allowed automatically. I shouldn't have to manually allow them.
3. Extremely obvious documentation around what blocks are allowed by default and where to change that.

ryanwelcher:
Looks like there will be some changes coming in 5.8 that affect this - https://make.wordpress.org/core/2021/06/16/block-editor-api-changes-to-support-multiple-admin-screens-in-wp-5-8/

PR #33, opened 2021-05-01
Remove DropCap setting from the Core Paragraph Block by default

fabiankaegy:
This will only work for WordPress 5.6 & 5.7 and is likely going to need an update for 5.8 again.

### Description of the Change

With this PR I am removing the DropCap option from the sidebar of the `core/paragraph` block.

The DropCap setting can get disabled via the theme.json file. Since this mechanism currently is only available in the Gutenberg Plugin there is a way to filter these setting in core. Since the theme.json format is still experimental however the format for disabling the dropCap option changed between WordPress 5.6 & 5.7. In 5.8 the theme.json will most likely be supported and this won't be necessary any longer.

### Checklist:

- [X] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [X] I have updated the documentation accordingly.

### Changelog Entry

Removed DropCap option of Core Paragraph block by default.

### DIFF
--- a/themes/10up-theme/includes/blocks.php
+++ b/themes/10up-theme/includes/blocks.php
@@ -28,6 +28,8 @@ function setup() {
 
 	add_action( 'init', $n( 'block_patterns_and_categories') );
 
+	add_action( 'init', $n( 'disable_drop_cap_setting' ) );
+
 	/*
 	If you are using the block library, remove the blocks you don't need.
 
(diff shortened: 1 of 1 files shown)


smy315:
@fabiankaegy since some projects will still need or want the DropCap support, I would be hesitant to include this in the Scaffold. Also, I think we'll get in the rabbit hole of bloating the Scaffold with code disabling a bunch of settings that aren't really useful. 

I think this code snippet is valuable and should be available somewhere but not baked into the Scaffold. Again main concern being that we can potentially bloat the Scaffold mainly because Gutenberg has lot of settings that aren't really useful. I personally can think of about 10 with border-radius setting for buttons on top of my list.

fabiankaegy:
@smy315 the reason I added this PR is because of a request that @joesnellpdx had made in a list of improvements to the scaffold. So we may want to check on what we actually want there :)

ryanwelcher:
@fabiankaegy the code looks fine to me. However, I am also a little hesitant to include this as the default state for the scaffold for the same reasons as @smy315. 

@joesnellpdx if there is a common list of features that we disable and it needs to be in the theme scaffold, I wonder if we can handle this with a custom filter like `tenup_disable_block_features` that will target those items specifically? It will bloat the scaffold a bit but if it's being done on the majority of projects, we'll save some engineering cycles so IMO it's worth it.

ryanwelcher (commented):
@fabiankaegy the code looks fine to me. However, I am also a little hesitant to include this as the default state for the scaffold for the same reasons as @smy315. 

@joesnellpdx if there is a common list of features that we disable and it needs to be in the theme scaffold, I wonder if we can handle this with a custom filter like `tenup_disable_block_features` that will target those items specifically? It will bloat the scaffold a bit but if it's being done on the majority of projects, we'll save some engineering cycles so IMO it's worth it.

joesnellpdx:
@ryanwelcher maybe this is something worthy of us hoping on a call.

There are several things we are considering moving forward on the scaffold that will add in our overall project initial build efficiencies... things we do 80% or more of the time (see https://docs.google.com/spreadsheets/d/1-aatYPaTtzG5kk0ik3FcxUaYL5xjIuGfchopDINU-Qo/edit#gid=0)

One of those items on the list is removing drop caps - hardly ever in designs and hardly ever used. 

Any of these updates _should_ be easily removed when they are not applicable to a particlar project.

So, maybe a function setting makes more sense `tenup_disable_dropcaps( 'true' );` ... or something similar. 

CC: @fabiankaegy

joesnellpdx:
@fabiankaegy 

Will you revisit this and the validity since the 5.8 release and any thoughts on moving this forward? 

I think the proper solution now would be a default theme.json with dropcaps removed from paragraphs.

```
{
  "version": 1,
  "settings": {
      "typography": {
        "dropCap": false
      }
  }
}
```

If recommending this, we need to also make sure that adding a theme.json file by default will not cause any other issues

Thanks!

fabiankaegy:
@joesnellpdx This is a great point. I think this PR can get closed because it doesn't apply to WordPress 5.8. 

I would suggest that we do add `theme.json` support to the Theme scaffold because it makes a whole bunch of things easier. But there are a few things that will need to get considered. I am creating a ticket for this specifically.

PR #35, opened 2021-05-05
Replacing @10up/scripts with 10up-toolkit

nicholasio:


themes/10up-theme/package.json
```diff
@@ -2,26 +2,26 @@
...
   "devDependencies": {
-    "@10up/scripts": "^1.3.1"
+    "10up-toolkit": "^1.0.0"
   },
   "dependencies": {
     "normalize.css": "^8.0.1",
     "prop-types": "^15.7.2"
   },
-  "@10up/scripts": {
+  "10up-toolkit": {
```
nicholasio, line 24: Both `@10up/scripts` and `10up-toolkit` will work. `10up-toolkit` is the preferred way.

mu-plugins/10up-plugin/package.json
```diff
@@ -2,25 +2,25 @@
...
   },
   "devDependencies": {
-    "@10up/scripts": "^1.3.1"
+    "10up-toolkit": "^1.0.0"
   },
   "dependencies": {
     "prop-types": "^15.7.2"
   },
-  "@10up/scripts": {
+  "10up-toolkit": {
```
nicholasio, line 23: Both `@10up/scripts` and `10up-toolkit` will work. `10up-toolkit` is the preferred way.

themes/10up-theme/package.json
```diff
@@ -2,26 +2,26 @@
...
   "scripts": {
-    "start": "10up-scripts start",
-    "watch": "10up-scripts watch",
-    "build": "10up-scripts build",
-    "format-js": "10up-scripts format-js",
-    "lint-js": "10up-scripts lint-js",
-    "lint-style": "10up-scripts lint-style",
-    "test": "10up-scripts test-unit-jest",
+    "start": "10up-toolkit start",
+    "watch": "10up-toolkit watch",
```
nicholasio, line 6: The build command also supports a `--watch` flag now. So it's also possible to do `"watch": "npm run build -- --watch"`

ryanwelcher: @nicholasio what is the difference between `watch` and the existing `start` command?

nicholasio: none, they're the same. `10up-toolkit start` `'10up-toolkit build --watch` and `10up-toolkit watch` is the same command.

in the future however we can differentiate them, like the start command could also spin up a webpackdev-server if 10up-toolkit is being used outside of WP context.

PR #41, opened 2021-06-17
Removing package-lock from scaffold

nicholasio:
### Description of the Change

`package-lock.json` shouldn't be in the scaffold and should only be introduced by projects.

### Checklist:

- [ ] My code follows the code style of this project.


joesnellpdx:
@nicholasio is this something we can move forward? I agree. Need to resolve conflicts, but will continue to cause issues until we can remove from version control.

Issue #39, opened 2021-05-28
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js): TypeError: $ is not a function


sharazghouri:
**Describe the bug**
![image](https://user-images.githubusercontent.com/17900945/119991719-2be76b00-bfe3-11eb-9857-7b4b4add50a9.png)

**Steps to Reproduce**
npm install
npm install -dev
npm run start

**Expected behavior**

CSS file not compiling and giving an error you can see the screeshot. It must be compile smothly.**Screenshots**
![image](https://user-images.githubusercontent.com/17900945/119991719-2be76b00-bfe3-11eb-9857-7b4b4add50a9.png)

**Environment information**
 - Device: 
 - OS: windows
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

sharazghouri:
An error has gone if I add .babelrc.js with this configuration 

```
module.exports = {
  presets: [
    ['@babel/preset-env', {
      "useBuiltIns": "usage" // do I need this?
    }]
  ],
  "plugins": [
    [
        "@babel/plugin-transform-runtime",
        {
            "corejs": 3 // Refer to the official documentation
          }
    ]
  ]
};module.exports = {
  presets: [
    ['@b
```abel/preset-env', {
      "useBuiltIns": "usage" // do I need this?
    }]
  ],
  "plugins": [
    [
        "@babel/plugin-transform-runtime",
        {
            "corejs": 3 // Refer to the official documentation
          }
    ]
  ]
};

turtlepod:
same issue here.
also using windows

fabiankaegy:
This was a webpack error that was fix

PR #42, opened 2021-06-18
Add skip to content link

AlejandroPerezMartin:
This PR adds a skip to content link to improve accessibility as described in: https://www.w3.org/TR/WCAG20-TECHS/G1.html (Requirement in WCAG 2.0).

### Benefits

Improve accessibility for keyboard-only users.

### Verification Process

- The _Skip to content_ link should be invisible until it gets focused.
- Press `Tab` and the link should now be visible.
- Pressing enter or clicking that button should change the focus to the main content area (#39 

### Checklist:

### Changelog Entry

ADDED: Skip to content link

### DIFF
--- a/themes/10up-theme/assets/css/frontend/base/index.css
+++ b/themes/10up-theme/assets/css/frontend/base/index.css
@@ -1,2 +1,3 @@
 @import url("prefers-reduced-motion.css");
 @import url("wordpress.css");
+@import url("utils.css");
--- a/themes/10up-theme/assets/css/frontend/base/utils.css
+++ b/themes/10up-theme/assets/css/frontend/base/utils.css
@@ -1,0 +1,33 @@
+/**
+ * Text meant only for screen readers
+ * Source: https://make.wordpress.org/accessibility/handbook/markup/the-css-class-screen-reader-text/
+ */
+.screen-reader-text {
+	border: 0;
+	clip: rect(1px, 1px, 1px, 1px);
+	clip-path: inset(50%);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
+	word-wrap: normal !important;
+
+	&:focus {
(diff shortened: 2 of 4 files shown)


AlejandroPerezMartin:
This is pending https://github.com/10up/wp-scaffold/pull/43 as the values may change depending on the base font size

dmtrmrv:
Hey @AlejandroPerezMartin @joesnellpdx let me propose a few changes to this PR:

1. I was never a fan of the class name `screen-reader-text` as it is a bit misleading. Technically all text on the screen that's not hidden from screen readers is a screen reader text. What we are doing with `screen-reader-text` is kind of the opposite, we are visually hiding an element but making it accessible to screen readers. With that said, I think we should call it `visually-hidden`.

2. I also think we should introduce two utility classes instead of one: `visually-hidden` and `visually-hidden-focusable`. First one does not require a focused state and is useful for things like:

```
<a href="#">Learn more <span class="visually-hidden">about {{articleTitlte}}</span></a>
```
... and the second for things like the _skip to content_ link.

3. Speaking of the _skip to content link_. I think it should be a separate component in CSS. `screen-reader-text` (or `visually-hidden-focusable`) class is meant to be used as a utility class, so applying styles meant only for _skip to content link_ to its focused state isn't correct. Most likely other elements with this class will need different position, font size, background and padding. We'll have to override those every time we use this class, which isn't productive.

For reference, I made proposed changes in this branch: https://github.com/10up/wp-scaffold/tree/feature/skip-to-content-link-updates

Feel free to compare these branches and let me know if you have any questions. Thanks!

AlejandroPerezMartin:
All your points makes sense to me @dmtrmrv ! Could you please open a new PR? Then I'll close this one! Thanks a lot! 😄

cc @joesnellpdx

dmtrmrv:
@AlejandroPerezMartin the new PR is here #47 We can close this one.

AlejandroPerezMartin:
thanks @dmtrmrv!

PR #43, opened 2021-06-21
Feature/font variables

AlejandroPerezMartin:
### Description of the Change

This PR adds CSS font-size and line-height variables as well helper classes. It also adds these font-sizes to the editor.

### Verification Process

- Open the editor and add a paragraph block, list of custom font sizes should be available: 

Screenshot 2021-06-21 at 13 42 42

### Checklist:

### Changelog Entry

- ADDED: Body and headings default font styles and helper classes
- ADDED: Font-size CSS variables
- ADDED: Custom font sizes to WordPress editor

### DIFF
--- a/themes/10up-theme/assets/css/frontend/global/typography.css
+++ b/themes/10up-theme/assets/css/frontend/global/typography.css
@@ -3,8 +3,8 @@
  */
 
 :root {
-	--ff-body: system-ui;
-	--ff-heading: system-ui;
+	--font-family-body: system-ui;
+	--font-family-heading: system-ui;
 
 	/*
 	 * Font Sizes
@@ -23,16 +23,16 @@
 	/*
 	 * Font Weights
 	 */
-	--fw-light: 300;
-	--fw-normal: 400;
-	--fw-medium: 500;
-	--fw-bold: 700;
+	--font-weight-light: 300;
+	--font-weight-normal: 400;
+	--font-weight-medium: 500;
+	--font-weight-bold: 700;
 
 	/*
 	 * Line Height
 	 */
-	--lh-small: 1.3;
-	--lh-base: 1.5;
-	--lh-medium: 1.6;
-	--lh-large: 1.7;
+	--line-height-small: 1.3;
+	--line-height-base: 1.5;
+	--line-height-medium: 1.6;
+	--line-height-large: 1.7;
 }
--- a/themes/10up-theme/assets/css/frontend/global/typography.css
+++ b/themes/10up-theme/assets/css/frontend/global/typography.css
@@ -9,16 +9,16 @@
 	/*
 	 * Font Sizes
 	 */
-	--fs-xxsmall: 13px;
-	--fs-xsmall: 14px;
-	--fs-small: 16px;
-	--fs-base: 18px;
-	--fs-medium: 22px;
-	--fs-xmedium: 26px;
-	--fs-xxmedium: 30px;
-	--fs-large: 38px;
-	--fs-xlarge: 48px;
-	--fs-xxlarge: 64px;
+	--fs-xxsmall: 0.8125rem;
+	--fs-xsmall: 0.875rem;
+	--fs-small: 1rem;
+	--fs-base: 1.125rem;
+	--fs-medium: 1.375rem;
+	--fs-xmedium: 1.625rem;
+	--fs-xxmedium: 1.875rem;
+	--fs-large: 2.375rem;
+	--fs-xlarge: 3rem;
+	--fs-xxlarge: 4rem;
 
 	/*
 	 * Font Weights
--- a/themes/10up-theme/assets/css/frontend/global/typography.css
+++ b/themes/10up-theme/assets/css/frontend/global/typography.css
@@ -25,7 +25,7 @@
 	 */
 	--fw-light: 300;
 	--fw-normal: 400;
-	--fw-medium: 400;
+	--fw-medium: 500;
 	--fw-bold: 700;
 
 	/*
--- a/themes/10up-theme/assets/css/frontend/global/typography.css
+++ b/themes/10up-theme/assets/css/frontend/global/typography.css
@@ -31,8 +31,8 @@
 	/*
 	 * Line Height
 	 */
-	--lh-default: 1;
-	--lh-base: 1.33;
-	--lh-medium: 1.4;
-	--lh-large: 1.6;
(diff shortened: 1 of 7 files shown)


AlejandroPerezMartin:
I have not worked yet on many 10up projects so those values might not be the most used ones, I took the ones that are present in the styleguide.
I'm also using PX instead of REM units cause I saw this https://github.com/10up/wp-scaffold/pull/20, but I can quickly convert that to REM if needed

barneyjeffries:
@AlejandroPerezMartin Joe asked me to take a look over the PR to help him out.

This is looking good. My preference for var naming is to use long-form - I favour --font-size-large over --fs-large - more verbose, easy to search etc. I would be interested to see what other devs think.

I definitely think fonts sizes should be in REMs - in discussion with @joesnellpdx yesterday he advised that new WP version might make the point raised here (https://github.com/10up/wp-scaffold/pull/20) not relevant.

Can we avoid using Extra Extra Medium? Perhaps instead use 2xl, 3xl and 2xs and 3xs so only need one medium?

And the font-weight medium should be 500 not 400.

cc @joesnellpdx

themes/10up-theme/assets/css/frontend/global/typography.css
```diff
@@ -0,0 +1,38 @@
...
+	--fs-large: 38px;
+	--fs-xlarge: 48px;
+	--fs-xxlarge: 64px;
+
+	/*
+	 * Font Weights
+	 */
+	--fw-light: 300;
+	--fw-normal: 400;
+	--fw-medium: 400;
```
barneyjeffries: @AlejandroPerezMartin This should be 500

themes/10up-theme/assets/css/frontend/global/typography.css
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+:root {
+	--ff-body: system-ui;
+	--ff-heading: system-ui;
+
+	/*
+	 * Font Sizes
+	 */
+	--fs-xxsmall: 13px;
```
barneyjeffries: @AlejandroPerezMartin consider using --font-size

themes/10up-theme/includes/core.php
```diff
@@ -60,6 +60,72 @@ function theme_setup() {
 		)
 	);
 
+	add_theme_support(
+		'editor-font-sizes',
+		array(
+			array(
+				'name'      => __( 'Extra Extra Small', 'tenup-theme' ),
```
barneyjeffries: @AlejandroPerezMartin Can we use 2x small or something like that to avoid "extra extra"?

themes/10up-theme/assets/css/frontend/global/typography.css
AlejandroPerezMartin: nice catch, thanks!

themes/10up-theme/includes/core.php
AlejandroPerezMartin: sure, renamed!

AlejandroPerezMartin:
@barneyjeffries both [color variables](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/assets/css/frontend/global/colors.css) and [breakpoints](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/assets/css/frontend/global/media-queries.css) are using short names, should I also update the naming there?

themes/10up-theme/assets/css/frontend/global/typography.css
AlejandroPerezMartin: @barneyjeffries both [color variables](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/assets/css/frontend/global/colors.css) and [breakpoints](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/assets/css/frontend/global/media-queries.css) are using short names, should I also update the naming there?

AlejandroPerezMartin:
> @AlejandroPerezMartin Joe asked me to take a look over the PR to help him out.
> 
> This is looking good. My preference for var naming is to use long-form - I favour --font-size-large over --fs-large - more verbose, easy to search etc. I would be interested to see what other devs think.
> 
> I definitely think fonts sizes should be in REMs - in discussion with @joesnellpdx yesterday he advised that new WP version might make the point raised here (#20) not relevant.
> 
> Can we avoid using Extra Extra Medium? Perhaps instead use 2xl, 3xl and 2xs and 3xs so only need one medium?
> 
> And the font-weight medium should be 500 not 400.
> 
> cc @joesnellpdx

pushed some changes, ready for you to review again! thanks! 🙂 Let me know if you still want to change the names, I updated `Extra Extra Large` to be `2 Extra Large` in the editor, maybe `2XL` is enough?

barneyjeffries:
Thanks @AlejandroPerezMartin - this looks good!

devinle:
Hey, @AlejandroPerezMartin the only issue I have is literally spelling out the px sizes for each of the headings in the page-styleguide. I realize this isn't on you, it was there previously, and so, this might need a different ticket.

Otherwise, as @barneyjeffries mentions, REM for the win for sure. I like em for line-height, but not for font sizing myself too.

cc @joesnellpdx

joesnellpdx:
@devinle @AlejandroPerezMartin @barneyjeffries 

We have to be careful with rems. They don't play nice in the block editor unless we commit to using the browser default font sizing for the html element. 

Per [the comment above](https://github.com/10up/wp-scaffold/pull/43#issuecomment-884710118), there is a future where the block editor lives within an iframe - which would be ideal, but I want to be clear on our intention. 

IF we go with rems for now, should we add commenting / guidance to ensure we don't mess with the html font sizes? 

====

Why would we change html font sizes?

1. Set base font size to 10px - so 1rem - 10px - easy math everywhere (personal preference, but not in the cards ATM)
2. Change overall spacing / sizing at certain breakpoints as a global change - REM is 16px by default, over 1200px wide viewport -  up to 18px (changes all the things in one line of css).

^^ for reference

rdimascio:
We should probably use `theme.json` for adding font sizes instead of `add_theme_support('editor-font-sizes')`, right?

joesnellpdx:
> We should probably use `theme.json` for adding font sizes instead of `add_theme_support('editor-font-sizes')`, right?

@devinle we should approach this the same as we do [here](https://github.com/10up/wp-scaffold/pull/44#issuecomment-1064343734)

PR #45, opened 2021-06-25
add debug plugins, wpackagist as a composer repository, gitignoring the plugin dir

tylercherpak:
### Description of the Change
Adding debug plugins to `require-dev` section of composer.json so that they are there by default for debugging and performance monitoring locally. You can avoid adding them by running `composer install --no-dev` instead of `composer install`

Gitignoring the plugin directory since it should be managed by composer. 

Adding wpackagist as a composer repository by default.

### Benefits
Better monitoring of performance and easier debugging.

### Checklist:

### Changelog Entry

Add debug plugins, wpackagist as a composer repository, gitignoring the plugin dir


nicholasio (approved):
Thoughts on activating this via code? Something like how VIP handles plugin activation via code.

We could have a mu-plugin for local stuff  `local-dev.php`, then only run the activation code if we're on local env (https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/)  and the plugin exists.

Maybe that's too much but would be a nice workflow improvement and would avoid us having to use wp-cli or manually active these plugins.

tylercherpak:
@nicholasio this would mean the code is on the production server but not being used. I would prefer we didn't have dead code on the production server.

PR #47, opened 2021-08-02
Add `visually-hidden` utility class and Skip To Content link

dmtrmrv:
### Description of the Change

This PR adds support for `visually-hidden` and `visually-hidden-focusable` utility classes and "Skip to content link". It also replaces this PR: https://github.com/10up/wp-scaffold/pull/42

Using `visually-hidden` and not `screen-reader-text` for a few reasons:

- WordPress 5.8 also has the `screen-reader-text` in `common.css` which is loaded on the front end if you use default block styles. That would be ok, but the core class has opinionated `:focus` styles. https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-admin/css/common.css#L146-L164 This means we'll have to override it in a theme, which isn't the cleanest approach.
- `visually-hidden` arguably seems to be a standard in a FE world:
  - https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/
  - https://www.sarasoueidan.com/blog/accessible-icon-buttons/
  - https://getbootstrap.com/docs/5.0/helpers/visually-hidden/

### Verification Process

- Verified that all elements with `.visually-hidden` class are hidden visually but announced by screen readers.
- Verified that all elements with `.visually-hidden-focusable` class are hidden visually but announced by screen readers and visible when they receive focus.

### Checklist:

### DIFF
--- a/themes/10up-theme/assets/css/frontend/base/index.css
+++ b/themes/10up-theme/assets/css/frontend/base/index.css
@@ -1,2 +1,3 @@
 @import url("prefers-reduced-motion.css");
 @import url("wordpress.css");
+@import url("utils.css");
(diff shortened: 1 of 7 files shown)


devinle:
@dmtrmrv this looks good to me! cc @joesnellpdx

PR #61, opened 2021-11-12
Added CT.CSS

### LINKED ISSUE #59: Add ct.css
Harry Roberts recently released a great little CSS utility that helps identify performance issues on sites called `ct.css`.
It is a diagnostic CSS stylesheet that looks something like this: https://csswizardry.com/ct/test/

This has a ton of benefits for adopting a performance culture at 10up where engineers can see with little effort what's being blocked or what can be optimized without adding to JS bloat.

My suggestion here would be to conditionally render `ct.css` behind a query string, say: `?debug_perf=1` which will trigger the stylesheets load.

The goal here is to quickly be able to diagnose what's blocking the rendering pipeline and solve it efficiently.

Great article at Smashing Magazine about this: https://www.smashingmagazine.com/2021/09/css-head-tag/

dainemawer:
### Description of the Change

This update adds an inline style tag to the head that, once requested with `debug_perf=1` - displays a visual representation to engineers of problematic scripts that are rendering resources / creating performance bottlenecks.

### Benefits

This change will allow engineers at any stage of the development process: development, staging, production to be able to determine what assets are rendering on page load. It has no JS dependency and is only loaded if the page is requested with `debug_perf=1`

### Verification Process

1. Validated inline CSS `<style>` in head
2. Refresh the page with `?debug_perf=1` - warnings displayed
3. Refresh the page without `?debug_perf` - no warnings displayed

### Checklist:

### Applicable Issues

 https://github.com/10up/wp-scaffold/issues/59

### Changelog Entry

[ ] Added [ct.css](https://csswizardry.com/ct/) to head, using query param in order to debug performance

### DIFF
--- a/themes/10up-theme/README.md
+++ b/themes/10up-theme/README.md
@@ -96,3 +96,13 @@ For VSCode you can add the following to your Settings. But other editors also su
 </details>
 
 <sub>* for 10uppers, reach out to Fabian for any questions / guidance / support in regards to `theme.json`</sub>
+
+# Performance Utilities
+The theme now supports `ct.css`. Uh what?
+`ct.css` is a diagnostic stylesheet that exposes potential performance issues in your pages `<head>` element. `ct.css` will return color-coded visual cues with regards to render blocking elements in the theme. This provides a great way for engineers to debug and identify problem resources.
+
+You can activate `ct.css` on any page load by including `?debug_perf=1` in the URL.
+
+Considering we do not want to load script everywhere throughout the theme, we have provided engineeers with a way to trigger the `ct.css` output by using a query param.
+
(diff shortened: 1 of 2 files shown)


joesnellpdx:
@dainemawer 

This looks great! Let's update a README to provide information guidance for usage. 

Suggestion:
We can add a section to the root README, or start a theme based README (I feel the latter is more appropriate). If we do the theme based one, use the root README format as a guide (include the readme title and theme description - can be generic - with a section for something like "Theme Utilities" or something like that). 

What do you think?

dainemawer:
@joesnellpdx README updated!

joesnellpdx (approved):

PR #69, opened 2022-01-11
Fix how styles are being loaded in the editor

fabiankaegy:
### Description of the Change

This PR changes how styles are being loaded in the editor. Instead of using the `enqueue_block_editor_assets` hook the `add_editor_style` function is used to enqueue the main stylesheet in the editor. This automatically takes care of wrapping everything to the `.editor-styles-wrapper` 

If there are things that need to style anything outside of the `.editor-styles-wrapper` you can still use the `editor-style-overrides.css` file. 

### Benefits

This loads all of the frontend styles in the editor also. So in theory if your markup matches you will only have to import the block stylesheet once into the frontend stylesheet. And core takes care of scoping everything to the editor styles wrapper.

Additionally the `add_editor_style` function also takes care of loading the stylesheet in all of the `iframed` editors like the site & template editors from 5.8 & 5.9

### Checklist:

### Changelog Entry

Changed how stylesheets are being loaded in the editor to also load the stylesheet in any framed editors like the site and template editors.


fabiankaegy:
@nicholasio with this change we can also remove the need for the post css plugin that scopes the editor stylesheet to the `.editor-styles-wrapper` since core is doing that for us.

nicholasio (approved):
👍 for this.

But I'll defer to @tlovett1 and @joesnellpdx to merge this.

joesnellpdx (approved):
Approved by me. @tlovett1 if you don't have any issues/questions, I'll go ahead and merge.

fabiankaegy:
@tlovett1 After our conversation yesterday I looked into the cash-busting nature of this and found that this actually isn't an issue here since the editor stylesheet gets inlined in the various editor contexts. 

From my perspective this is ready to be merged.

nicholasio:
Would also note that there are currently a few issues with the postcss editor-style plugin that aren't easily fixable and would require forking 

Once we get this merged I'd also like to get rid of `postcss-editor-style` in a future major version of 10up-toolkit.

tlovett1:
Assuming this has been thoroughly tested, I'm comfortable with someone merging it

PR #78, opened 2022-01-24
Update Block Registration to Happen Automatically for any blocks containing a block.json file

### LINKED ISSUE #62: Simplify & standardize how to add custom blocks to the theme scaffold
**Is your enhancement related to a problem? Please describe.**

Currently, the way to add a new block is rather cumbersome. You have to first create a new JS entry point in the `package.json` file like so: 

```json
{
  "10up-toolkit": {
    "entry": {
      "example-block": "./includes/blocks/example-block/index.js"
    }
  }
}
``` 

Next you have to require the PHP file like so: 

```php
require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';

// Call block register functions for each block.
Example\register();
```

I think both of these steps could be skipped if we are able to automatically create an entry point for every folder inside the `blocks` directory and also use autoloader to load all the PHP `register.php` files inside the individual block folders.

fabiankaegy:
### Description of the Change

This changes the block registration flow so that all blocks located within the directory that have a `block.json` file get automatically registered.

They also automatically get their  added if the directory where the  file is located also contains a  file

Closes #62 

### Checklist:

### Credits

Props @


themes/10up-theme/includes/blocks.php
```diff
@@ -42,19 +42,48 @@ function setup() {
...
 	// Filter the plugins URL to allow us to have blocks in themes with linked assets. i.e editorScripts
 	add_filter( 'plugins_url', __NAMESPACE__ . '\filter_plugins_url', 10, 2 );
 
-	// Require custom blocks.
-	require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';
+	// get all block.json files within the includes/blocks folder
+	$old_working_dir = getcwd();
+	chdir( ABSPATH );
+	$block_json_files = glob( 'wp-content/themes/10up-theme/includes/blocks/*/block.json' );
+	chdir( $old_working_dir );
```
fabiankaegy: @tlovett1 This feels super hacky to me. Is there a better way of doing this?

joesnellpdx (commented):
Besides the open comment left for @tlovett1 , this is approved by me.

fabiankaegy:
@gthayer I have updated this PR based on how #85 includes all the files. In my testing, this still works and feels much lass hacky. Let me know what you think.

themes/10up-theme/includes/blocks.php
```diff
@@ -42,19 +42,49 @@ function setup() {
...
 	add_filter( 'plugins_url', __NAMESPACE__ . '\filter_plugins_url', 10, 2 );
 
-	// Require custom blocks.
-	require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';
+	// Register blocks for the active theme, for both parent and child theme,
+	// if applicable.
+	foreach ( wp_get_active_and_valid_themes() as $theme ) {
+		$block_directory_path = $theme . '/includes/blocks/';
+		if ( file_exists( $block_directory_path ) ) {
+			$block_json_files = glob( $block_directory_path . '*/block.json' );
```
gthayer: Is there a reason you are attempting to read the locations instead of using the constants defined in the theme?

https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/functions.php#L15

themes/10up-theme/includes/blocks.php
```diff
@@ -42,19 +42,49 @@ function setup() {
...
 function register_theme_blocks() {
 	// Filter the plugins URL to allow us to have blocks in themes with linked assets. i.e editorScripts
 	add_filter( 'plugins_url', __NAMESPACE__ . '\filter_plugins_url', 10, 2 );
 
-	// Require custom blocks.
-	require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';
+	// Register blocks for the active theme, for both parent and child theme,
+	// if applicable.
+	foreach ( wp_get_active_and_valid_themes() as $theme ) {
+		$block_directory_path = $theme . '/includes/blocks/';
```
gthayer: I think this is the same as `TENUP_THEME_BLOCK_DIR`

https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/functions.php#L15

themes/10up-theme/includes/blocks.php
fabiankaegy: The only reason there is that right here we are looping over child themes as well. 

This is why it is being build out of the theme loop here:
```php
foreach ( wp_get_active_and_valid_themes() as $theme ) {
```

themes/10up-theme/includes/blocks.php
fabiankaegy: Yes but with the additional support of child themes. 

(which might not be super important)

themes/10up-theme/includes/blocks.php
```diff
@@ -42,19 +42,49 @@ function setup() {
...
  */
 function register_theme_blocks() {
 	// Filter the plugins URL to allow us to have blocks in themes with linked assets. i.e editorScripts
 	add_filter( 'plugins_url', __NAMESPACE__ . '\filter_plugins_url', 10, 2 );
 
-	// Require custom blocks.
-	require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';
+	// Register blocks for the active theme, for both parent and child theme,
+	// if applicable.
+	foreach ( wp_get_active_and_valid_themes() as $theme ) {
```
gthayer: Noting that `wp_get_active_and_valid_themes()` is for core use only.

https://developer.wordpress.org/reference/functions/wp_get_active_and_valid_themes/

fabiankaegy: Ressolved (y)

themes/10up-theme/includes/blocks.php
```diff
@@ -42,19 +42,46 @@ function setup() {
...
  * @return void
  */
 function register_theme_blocks() {
 	// Filter the plugins URL to allow us to have blocks in themes with linked assets. i.e editorScripts
 	add_filter( 'plugins_url', __NAMESPACE__ . '\filter_plugins_url', 10, 2 );
 
-	// Require custom blocks.
-	require_once TENUP_THEME_BLOCK_DIR . '/example-block/register.php';
+	// Register all the blocks in the theme
+	$block_directory_path = TENUP_THEME_BLOCK_DIR;
```
gthayer: Not a huge deal, but you could just be using the constant here. No need to declare it to a variable.

gthayer: Resolved.

gthayer:
LGTM!

PR #83, opened 2022-02-14
Removing lerna in favor of npm workspaces (npm >=7)

nicholasio:
### Description of the Change

Lerna was added to the scaffold to allow us to better handle npm dependencies in the new scaffold structure but, we're only using `lerna` to handle installing dependencies. However, npm workspaces have been introduced in npm 7 and we no longer need lerna in the scaffold.

Lerna also requires following a different workflow for adding deps (`lerna add`) which can be a bit cumbersome to unfamiliar engineers.

By moving to `npm workspaces` we reduce one dependency (which isn't very actively maintained at this point) and we simplify the overall workflow.

Instead of `lerna.json`, we juts need to add  `workspaces` in package.json.

Adding/remove deps is much more familiar to everyone as you only need to use a flag to specify the package:
`npm install --save prop-types -w=package-name`

Closes #

### Possible Drawbacks

npm workspaces can't run scripts in parallel so blocking scripts (like watch/start) will prevent npm running the rest of the commands. To workaround this, `npm-run-all` has been introduced.

### Verification Process

- Remove `node_modules`
- Make sure you're using npm >= 7
- npm install
- npm run build
- npm run watch
- npm run start

### Checklist:

- [x My code follows the code style of this project.

### Changelog Entry

Changed - Removing `lerna` in favor of `npm workspaces`

### Credits

Props @nicholasio


nicholasio:
@dustinrue @davidegreenwald Is managed hosting/system leveraging `lerna.json` for anything in the ci pipeline?

dustinrue:
@nicholasio no, CI system only ensured lerna could be run at all, everything else is dependent on build scripts

nicholasio:
Added `enginesStrict: true` to force the node & npm versions in package.json

image

tlovett1:
I'm good with this. However, I have the same request as I do for 10up-toolkit which is better error messaging. I'd love some louder error messaging when you try to npm install not meeting the minimum node/npm reqs.

nicholasio:
> I'm good with this. However, I have the same request as I do for 10up-toolkit which is better error messaging. I'd love some louder error messaging when you try to npm install not meeting the minimum node/npm reqs.

So with this option, running `npm install` won't work at all in the scaffold if you don't meet the minimum reqs.

re: 10up-toolkit. while 10up-toolkit had some issues with some specific node/npm versions in the past (which is being addressed in the upcoming 10up-toolkit release) 10up-toolkit works with node >=12 and npm >= 6. In general the "developer experience" is something that I plan on keeping improving on 10up-toolkit, so I will definitely improve error messages and such.

PR #86, opened 2022-02-28
Import & Run

### LINKED ISSUE #80: Add importAndRun function for Code Splitting

Based on a recent Slack conversation: https://10up.slack.com/archives/G11CA43FW/p1643728684014569
@rdimascio and I believe it would be a great idea to include this boilerplate JavaScript code into `frontend.js` so that engineers can better work with dynamically imported JavaScript components:

```
import importAndRun from './util/import-and-run';

const COMPONENTS = {
	'components/accordion-block': {
		selectors: ['.accordion'],
	},
	'components/site-header': {
		selectors: ['body'],
	},
	'components/site-footer': {
		selectors: ['body'],
	},
	'components/story-navigation': {
		selectors: ['.home-nav'],
	},
	'components/service-explorer': {
		selectors: ['.service-explorer'],
	},
	'components/story-comments': {
		selectors: ['.story-comments'],
	},
	'components/social-share': {
		selectors: ['.social-share'],
	},
	'components/child-filters': {
		selectors: ['.photolisting__aside-filters'],
	},
	'components/cookie-consent': {
		selectors: ['#cookie-consent'],
	},
};

const initComponents = () => {
	Object.entries(COMPONENTS).forEach(([key, value]) => {
		importAndRun(key, value.selectors, value.callback);
	});
};

const init = () => {
	window.requestAnimationFrame(() => {
		document.body.classList.add('js-loaded');
	});

	initComponents();
};

/* eslint-disable-next-line @wordpress/no-global-event-listener */
document.addEventListener('DOMContentLoaded', init);

```

The above allows us to cleanly and easily keep track of components and import / run them as required. 
This task would also require us adding the `importAndRun` utility to a new `assets/js/util` folder.

It's our hope that this will close the loop on dynamically imported JavaScript as the scaffold already supports Code Splitting out the box. 

```
/**
 * Default callback when using `importAndRun`
 *
 * @param {Object} Module - Factory function or class
 */
const DEFAULT_CALLBACK = (Module) => {
	try {
		const module = new Module();
		module.init();
	} catch (error) {
		// Something failed
	}
};

/**
 * Dynamically import a module if it appears on the page.
 *
 * @param {string} module - Path to module
 * @param {string[]} selectors - List of selectors. Used to conditionally load the module
 * @param {Function} cb - Callback function to run after loading the module
 * @return {Promise} - Returns the dynamic import
 */
export default (module, selectors = ['body'], cb = DEFAULT_CALLBACK) => {
	let shouldImport = false;

	if (Array.isArray(selectors) && selectors.length) {
		shouldImport = Boolean(
			selectors.find((selector) => {
				return Boolean(document.querySelector(selector));
			}),
		);
	}

	if (!shouldImport) {
		Promise.resolve();
	}

	/* eslint-disable-next-line consistent-return */
	return import(`./../${module}`).then(({ default: module }) => {
		if (cb && typeof cb === 'function') {
			cb(module);
		}
	});
};

```

### Designs

_No response_

_No response_

### Code of Conduct

dainemawer:
### Description of the Change

Fixes #80 

### Checklist:

### Credits

Props @


dainemawer:
@rdimascio - I noticed a strange bug here. The `navigation` component should not run unless the `selectors` array matches, but it seems to run regardless. Do you mind taking a look?

rdimascio:
Thanks for putting this together @dainemawer! I fixed the bug you mentioned, cleaned up the architecture a bit, added error handling and some documentation. This is looking good to me.

cc @joesnellpdx @devinle @nicholasio @fabiankaegy

rdimascio (approved):

devinle:
@rdimascio @dainemawer this is looking good to me. Nice and clean - easy to follow. I also like the ability to pass a custom callback. Well done!

joesnellpdx:
Per discussion with @rdimascio ... 

Is this something we want to include in 90% of our projects, or are there particular use cases that would trigger the usage of import-and-run?

If it is not something we want to have as default, do we consider one of the following?:

- Add as an enhancements section in the README - with guidance / examples
- Add a new enhancements doc - link to it in the README - and include guidance / examples for this
- Add an examples directory in the project that is not included with references in the README file / comments
- Add to existing architecture, but comment out with comment guidance when/ how to activate (and add to README)

Either way, there needs to be more clear guidance and documentation for engineers if this is implemented. 

@devinle @dainemawer

themes/10up-theme/assets/js/frontend/components/navigation.js
```diff
@@ -0,0 +1,5 @@
+export default class Navigation {
```
nicholasio, line 1: Does this actually need to be a class?

nicholasio (commented):
What are we thinking in terms of recommendations for this?

Are we going to essentially code split every single front-end component?

I do not think code-splitting every single front-end component is worth it. There's an overhead of making an additional call to load the JS. If we ever ship something like this I feel we need more guidance around when to use it. While I do not have any benchmark to back me up here I do not think code-splitting very small components would actually give any meaningful improvement, while potentially making things a little worse. We should also be mindful of CLS, if every single front-end component is lazy-loaded, this might mean we're delaying fully rendering stuff on the front end which could cause content shifting

rdimascio:
@nicholasio I would think that script parse and execution time on page load would result in a worse user experience and CWV scores than network requests. I don't have any benchmarks for this either (and @dainemawer can probably provide more insight here) but I believe the implementation for this approach (when and even how to use it) will larger depend on project architecture. I'm in agreement that there needs to be more clear guidance and documentation on this.

themes/10up-theme/assets/js/frontend/components/navigation.js
rdimascio: It does not, but I think the large majority of our projects are still using class based components. I added a [comment](https://github.com/10up/wp-scaffold/pull/86/files#diff-d35de96f686e5dbc10d19b9db2b20419ac339e0e5d2d04c8d3ea5618ecc6de90R23-R26) for sites with functional components.

nicholasio:
> xecution time on page load would result in a worse user experience and CWV scores than network requests. I don't have any benchmarks for this either (and @dainemawer can probably provide more insight here) but I believe the implementation for this approach (when and even how to use it) will larger depend on project architecture. I'm in agreement that there needs to be more clear guidance and documentation on this.

You're right but it depends on what you're code splitting. Every optimization comes with an overhead cost. Finding the balance is what we're missing here. We're all on the same page we need more documentation and guidance here 👍

themes/10up-theme/assets/js/frontend/components/navigation.js
dainemawer: @nicholasio this was just added for example sake - as long as this component is the default export we should be okay. Happy to update it to a `const Navigation = () => {}`

dainemawer:
> What are we thinking in terms of recommendations for this?
> 
> Are we going to essentially code split every single front-end component?
> 
> I do not think code-splitting every single front-end component is worth it. There's an overhead of making an additional call to load the JS. If we ever ship something like this I feel we need more guidance around when to use it. While I do not have any benchmark to back me up here I do not think code-splitting very small components would actually give any meaningful improvement, while potentially making things a little worse. We should also be mindful of CLS, if every single front-end component is lazy-loaded, this might mean we're delaying fully rendering stuff on the front end which could cause content shifting

@nicholasio - yeah I agree here, I don't think the intention is to code-split every component. What this PR lacks, as @joesnellpdx mentioned above, is documentation is going forward. The idea here should be to code-split out:

- Any resource heavy functions / components - potentially components requiring third-party libs like `lodash`, `moment` etc.
- Any scripts that do not need to be loaded in the bundle on initial load - chat libs, cookie banners etc.

I will add some guidance here. I'll also clean up this entry file so that its more instructional.

dainemawer:
In general, benchmarks are a bit all over the place. They are specific to the nature of the site as well. The overall strategy here is to offload any components and scripts that are not absolutely needed for a good user experience on initial load - or above the fold. 

An example of this would be something along the lines of:
1. We may write a component in JS for the Navigation. That JS should not be lazy-loaded. It's imperative for good UX on initial load.
2. A cookie banner though, whose presence is not in the flow of the document, and does not contribute to important user experience. 

Doing this above, and identifying elements with the above nature, means we can keep the byte size of our bundle low.

composer.json
```diff
@@ -17,5 +17,10 @@
   "scripts": {
     "lint": "phpcs .",
     "lint-fix": "phpcbf ."
+  },
+  "config": {
+    "allow-plugins": {
+      "dealerdirect/phpcodesniffer-composer-installer": true
```
tylercherpak, line 23: @dainemawer is this directly related to the dynamic imports?

tylercherpak (commented):
@dainemawer I love this and think it has a ton of potential to be part of how we think about components and if they are needed at time of initial paint. 

One question I want to raise for everyone, do we think we should have an example in which making the component dynamic is more necessary. Such as a slider that needs to include additional libraries that we don't want to load at render or if that component isn't present on the page. 

Let me know what you think! Thanks again for this!! 😄

themes/10up-theme/assets/js/frontend/components/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Initialize frontend components
+ *
+ * Run through the components map above and import them.
+ */
+export const initComponents = () => {
+	Object.entries(COMPONENTS).forEach(([component, args]) => {
+		importAndRun(
+			`frontend/components/${component}`,
```
tylercherpak, line 47: are we importing and running the un-minified js source? Should webpack be creating a separate bundle for this component?

fabiankaegy:
closing for now

PR #87, opened 2022-02-28
Feature/10up toolkit 3.0.0

nicholasio:
### Description of the Change

Closes #

This updates `wp-scaffold` to work with 10up-toolkit@3.0.0 and HMR & Fast Refresh.

It basically removes all css entry points from `package.json` and imports them via a corresponding js file.

### Verification Process

- Clone this branch.
- Make sure to remove node_modules and start from a clean state.
- Run `npm install` from the root
- Set SCRIPT_DEBUG to true
- `npm run watch`
- Edit any CSS file (admin-style.css, frontend.css, editor-styles-overrides.css, styleguide.css etc...) Changes should hot reload
- Edit the example block css and react code. Changes should hot reload.
- 
### Checklist:

### Credits

Props @


mu-plugins/10up-plugin/package.json
```diff
@@ -15,13 +15,12 @@
     "node": ">=12.0.0"
   },
   "devDependencies": {
-    "10up-toolkit": "^2.2.0-alpha.6"
+    "10up-toolkit": "^2.2.0-alpha.7"
```
fabiankaegy: I think this should only get merged once the actual 2.2.0 version has been released

themes/10up-theme/package.json
```diff
@@ -15,13 +15,12 @@
     "node": ">=12.0.0"
   },
   "devDependencies": {
-    "10up-toolkit": "^2.2.0-alpha.6"
+    "10up-toolkit": "^2.2.0-alpha.7"
```
fabiankaegy: Same here

mu-plugins/10up-plugin/package.json
```diff
@@ -15,20 +15,16 @@
     "node": ">=12.0.0"
   },
   "devDependencies": {
-    "10up-toolkit": "^2.1.0"
+    "10up-toolkit": "^2.2.0-alpha.8"
```
fabiankaegy: Same here

mu-plugins/10up-plugin/package.json
```diff
@@ -15,20 +15,16 @@
...
     "prop-types": "^15.7.2"
   },
   "10up-toolkit": {
-    "devURL": "https://my-project.test",
     "entry": {
       "admin": "./assets/js/admin/admin.js",
       "frontend": "./assets/js/frontend/frontend.js",
-      "shared": "./assets/js/shared/shared.js",
-      "admin-style": "./assets/css/admin/admin-style.css",
-      "shared-style": "./assets/css/shared/shared-style.css",
```
fabiankaegy, line 30: I would love to better understand how the entry points now get generated. Is is depending on the name of the JS file they were imported into? Or is it depending on the name of the stylesheet that gets imported? 

There will be cases where a block for example has two separate stylesheets. One for the editor and one for the fronend. They should not be merged into the same stylesheet. 

I think `@wordpress/scripts` handles this by allowing for `style.css` or `editor.css`. And depending on which it is they get put into two separate files.

Additionally, it would be great for the next iteration of all this to allow for per block stylesheet generation. That way there could be a `style.css` and `editor.css` for every single block.

themes/10up-theme/includes/blocks.php
```diff
@@ -113,6 +113,16 @@ function blocks_editor_styles() {
 		Utility\get_asset_info( 'editor-style-overrides', 'version' )
 	);
 
+	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
```
fabiankaegy, line 116: I could also see using `wp_get_environment_type( 'local' )` here.

themes/10up-theme/includes/blocks/example-block/edit.js
```diff
@@ -4,16 +4,19 @@
 import { __ } from '@wordpress/i18n';
 import { RichText, useBlockProps } from '@wordpress/block-editor';
 
+// Importing the block's css via JS will enable hot reloading for css
+import './index.css';
```
fabiankaegy: As part of this, we should rename the `index.css` file to `editor.css`. It currently is confusing and especially now with this new comment it makes it sound as if all the block styles should be written in that file.

themes/10up-theme/includes/core.php
```diff
@@ -104,6 +108,10 @@ function scripts() {
 		);
 	}
 
+	/**
+	 * Enqueuing shared.js is required to get css hot reloading working in the frontend
+	 * If you're not shipping any shared js wrap this enqueue in a SCRIPT_DEBUG check.
+	 */
 	/*
```
fabiankaegy, line 115: @nicholasio I think you forgot to uncomment this code here.

themes/10up-theme/package.json
```diff
@@ -15,24 +15,19 @@
     "node": ">=12.0.0"
   },
   "devDependencies": {
-    "10up-toolkit": "^2.1.0"
+    "10up-toolkit": "^2.2.0-alpha.8"
```
fabiankaegy: Again I think we should only merge with the final version.

fabiankaegy (changes requested):

mu-plugins/10up-plugin/package.json
nicholasio: Yep, I just wanted to get this reviewed ahead of final release.

mu-plugins/10up-plugin/package.json
nicholasio: The entrypoint gets generated based on the js file importing them.

So anything imported by `frontend.js` will get bundled in a `frontend.css` including block css etc.

I'm happy to expand this a bit further.

themes/10up-theme/includes/blocks/example-block/edit.js
nicholasio: Sure, I'll rename this to editor.css

themes/10up-theme/includes/core.php
nicholasio: So the shared style were already commented out in the scaffold so I left commented out.

mu-plugins/10up-plugin/package.json
fabiankaegy: Okay yeah, I think this is fine for now and makes sense. In the future when we want to tackle the individual stylesheets per block we will need to differentiate between editor and frontend styles. We can use `@wordpress/scripts` as an inspiration there.

themes/10up-theme/includes/blocks/example-block/edit.js
nicholasio: Done!

mu-plugins/10up-plugin/package.json
nicholasio: Sounds good.

We could expand this as part of the work to automatically detect block.json files in 10up-toolkit

themes/10up-theme/includes/blocks.php
nicholasio: That could work too, but HMR only works when SCRIPT_DEBUG is set to true and this enqueue only exists for the purpose of HMR ... so even if the env is set to local the SCRIPT_DEBUG constant could still be false.

joesnellpdx:
@nicholasio 

I tried to run a quick test - and maybe there is some user error, but here are my findings:

To test the toolkit 3.0.0 MR... how much setup do I need to get it to work?

- I removed all `node_modules` folders
-I have node 16.14.0 and npm 7.24.2
- I added the them devurl: "devURL": "https://10up-scaffold.test", (based on docs, using .test I may not need this .. so tested with and without)
- I'm running everything from the root
- I'm testing here `https://10up-scaffold.test/`  (I don't think I need a unique URL)

If I run `npm run watch` ... 
I change something in css... I see it compile... but have to refresh the page to see the change.
JS - files are not being enqueued (although I see `frontend.js` and `frontend.asset.php` in `/dist`)  so I can't test JS
Same is true if I run `npm run watch:theme`

^^ tried in the plugin as well - same results

FYI: there is still a `lerna clean` command in root `package.json` 

I don't have the bandwidth to debug currently, but wanted to give you an update. 

== 

I didn't test anything else.

nicholasio:
> @nicholasio
> 
> I tried to run a quick test - and maybe there is some user error, but here are my findings:
> 
> To test the toolkit 3.0.0 MR... how much setup do I need to get it to work?
> 
> * I removed all `node_modules` folders
> * I have node 16.14.0 and npm 7.24.2
> * I added the them devurl: "devURL": "https://10up-scaffold.test", (based on docs, using .test I may not need this .. so tested with and without)
> * I'm running everything from the root
> * I'm testing here `https://10up-scaffold.test/`  (I don't think I need a unique URL)
> 
> If I run `npm run watch` ... I change something in css... I see it compile... but have to refresh the page to see the change. JS - files are not being enqueued (although I see `frontend.js` and `frontend.asset.php` in `/dist`) so I can't test JS Same is true if I run `npm run watch:theme`
> 
> ^^ tried in the plugin as well - same results
> 
> FYI: there is still a `lerna clean` command in root `package.json`
> 
> I don't have the bandwidth to debug currently, but wanted to give you an update.
> 
> ==
> 
> I didn't test anything else.

Did you set SCRIPT_DEBUG to true?
`define('SCRIPT_DEBUG', true);`

joesnellpdx:
> Did you set SCRIPT_DEBUG to true? `define('SCRIPT_DEBUG', true);`

OMG! Dumb. That did it. Let take it for a deeper spin later today. Thanks!

joesnellpdx (approved):
@nicholasio 

Approved by me!

Only change maybe is removing `"clean": "lerna clean",` from the root `package.json`

nicholasio:
> @nicholasio
> 
> Approved by me!
> 
> Only change maybe is removing `"clean": "lerna clean",` from the root `package.json`

Done!

joesnellpdx:
@devinle @fabiankaegy 

Are either of. you able to take a pass by EOW? If not, I recommend @nicholasio go ahead and merge.

fabiankaegy (approved):
@joesnellpdx I've already given my approval when we tested it in a slack conversation  👍 this is approved from my end  👍

nicholasio:
Thanks everyone for the help reviewing and testing this!

PR #97, opened 2022-03-22
allow husky to install properly regardless of where wp-scaffold is used

dustinrue:
### Description of the Change

When wp-scaffold is used deeper within a git project, npm install/prepare will fail because the .git directory is not located with package.json. This PR changes the "prepare" stage of npm install so that husky is always installed properly. It also adjusts the pre-commit script so that it can properly find the files it is responsible for again.

Closes #

### Possible Drawbacks

The changes use bash commands, these may or may not be available in all environments but I'm not sure what the requirements or expectations are for the scaffold.

### Verification Process

Tested on macOS and CentOS 7 with wp-scaffold being in a child directory away from .git as well as with .git.

### Checklist:

### Changelog Entry

Changed - allow husky to be installed/used regardless of where wp-scaffold exists in the git repository

### Credits

Props @dustinrue

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,11 +1,10 @@
 #!/bin/bash
 . "$(dirname "$0")/_/husky.sh"
 
-PARENT_DIR=$(git rev-parse --show-toplevel)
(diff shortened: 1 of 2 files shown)


darylldoyle (approved):
Tested and this is working great!

PR #103, opened 2022-04-07
Fix: Block registrations in Themes in WordPress 6.0 by removing the filter of `plugins_url`

### LINKED ISSUE #99: Block registration from within the theme is broken in upcoming WordPress release
In my testing for the upcoming WordPress 6.0 release at the end of May I've found an issue with how we register blocks within the theme scaffold. In order to even be able to register bocks from within a theme, a workaround was added here: https://github.com/10up/wp-scaffold/blob/a2586e31dcea350faa6d28373e606c17a045d44d/themes/10up-theme/includes/blocks.php#L51 to override the `plugins_url` value with the theme url. 

This worked just fine with all the previous WordPress versions before 6.0 but with some underlying changes (I think this here: https://github.com/WordPress/wordpress-develop/commit/f034bc832eb091079e4130f6a206b4f496766796#diff-8c99af92e4ec0fdb307ddd9b42be1e1ef1efe4a9f31287c23f346244dddd1ce9R1210-R1246) the filter now breaks some core asset paths. 

In order to resolve this issue we either need to somehow modify the filter we are adding to the `plugins_url` to only alter the path when the path is for the block we are trying to enqueue, and not for any paths. 

Or we need to get WordPress core to allow for the inclusion of custom blocks from within themes.

fabiankaegy:
This patch is what I would recommend to any project that registers blocks with `block.json` files in a theme. The patch can be applied at any time before updating to WordPress 6.0. So we can get ahead of the actual update. 

The need for this fix was outlined in #99  

Since the patch was just merged into core today and we don't jet have a beta version of 6.0 to be more confident of the patch actually landing in there I am keeping this as a draft PR.

Closes #99


joesnellpdx (approved):
Approved when ready / after we have provided guidance to the team (next week or two).

fabiankaegy:
I have tested this today with the actual Beta 1 of WordPress 6.0 and can confirm that it is working as intended  👍 

I would recommend to hold till Beta 2 to ensure that the patch does not get removed before we merge this in and inform the team. That would be on April 19.

fabiankaegy:
Just to provide an update here. I've also tested Beta 2 & Beta 3 and the patch is still working as intended.

PR #107, opened 2022-04-13
automatically enqueue block frontend assets when block is used

### LINKED ISSUE #106: Automatically enqueue frontend assets of blocks defined in `block.json` for dynamic blocks
WordPress Core automatically enqueues frontend assets like the `viewScript` only on pages where the block is actively being used for static blocks. Dynamic blocks currently need to handle that step manually. 

### Proposed Solution:

Since WordPress already registers all the assets we can add a check whether an asset exists and then automatically enqueue it from within the `render_callback` function.

fabiankaegy:
### Description of the Change

Automatically enqueue registered assets when a block is being used on the page.

When a block defined an asset for its frontend usage (`style`, `script`, `viewScript`), that asset will automatically get registered by WordPress under the hood using the naming scheme `${namespace}-${blockName}-${assetName}`. So a view script of a block called `tenup/example` would get registered with the handle `tenup-exampe-view-script`.

When you are writing a static block (storing its markup in the database) these frontend assets automatically get enqueued on a page if the block is being used on that page.

This does not happen automatically for Dynamic blocks (PHP rendering). In a Dynamic block, you as the block author need to manually enqueue the registered asset via its handle. This is not really obvious since you have to know what the handle of the registered asset is going to be before you can even enqueue it. 

Because of that, this PR adds the same behavior that static blocks get from core to dynamic blocks out of the box.

Closes #106 

### Possible Drawbacks

There can be circumstances where you may only want to actually enqueue a view script for example if certain attributes of the block are present. You get that level of fine grain control when you manually handle the enqueueing. But if we handle it automatically there is no fine grain control. As soon as the block gets used on the page the relevant assets will get enqueued. 

### Checklist:

### Credits

Props @


themes/10up-theme/includes/blocks.php
```diff
@@ -71,6 +71,20 @@ function register_theme_blocks() {
 					$context            = $block->context;
 					$wrapper_attributes = wp_kses_post( get_block_wrapper_attributes() );
 
+					$frontend_assets = [
+						'script',
+						'viewScript',
+						'style',
+					];
+
+					foreach ( $frontend_assets as $asset_name ) {
```
fabiankaegy, line 79: This will also need to handle arrays and already provided handles.

the `viewScript` key can either be a relative file path `file:./...`, a script handle `my-handle` or an array containing both of these values multiple times.

All assets should get enqueued.

fabiankaegy:
This is no longer required because WordPress Core actually changed it's behavior and is now also automatically enqueueing all assets that are assigned to a block for dynamic blocks on the frontend 🎉

PR #108, opened 2022-04-14
Fix: Escaping of block wrapper attributes function

fabiankaegy:
Removes the incorrect triple escaping of the `get_block_wrapper_attributes` function

PR #110, opened 2022-04-23
WIP: feature/10upui-plugin

devinle:
**WIP**
The following PR is a first pass at updating our scaffold to work with the up and coming UI plugin. This is very early stages, and includes several updates to work across the theme to the plugin. 

**Related Repo:** 
https://github.com/10up/wp-scaffold-ui 

**Related Branch:** 
https://github.com/10up/wp-scaffold-ui/pull/9

**Notes**
1. Activate the 10up theme in the main WP Scaffold repo
2. Clone the plugin WP Scaffold UI and activate in the plugins directory
3. In the plugins/wp-scaffold-ui directory, you will notice a components directory. This is where we will store each new component we create. Each component would have an index.php file where the output lives. Along with a src directory for any CSS or JS work.
4. The CSS will be set to output into the dist directory as a standalone CSS prod file. This, so an Engineer can choose individual files to include in their theme, rather than a full stylesheet.
5. CSS classes should be pluggable by theme.json. In the case of the button, we define each CSS declaration using the main theme.json property name. Secondary to this, we also provide a fallback to a foundational setting stored within the plugin file itself. This foundational style is intentionally branded to align with 10up. This provides a fallback to quickly prototype, before adding any theme specific CSS settings.
6. Also in the plugin are global tokens. You can find these here: assets/css/frontend/tokens/*.css. These tokens are all prefixed with —tui (10upUI) for now. Happy to discuss this further.
7. The theme.json settings I’ve provided are trivial. Simply for testing and presentation only. We’d likely start by defining a palette–the core colors our theme would be using.
8. From there, we can map to align with our plugin’s button component as you will see under the custom object in the theme.json. These settings will be parsed and will align with our primary CSS var declarations made in the plugin button.css file. And, as mentioned previously, if not provided via theme.json, it will fallback to our plugin’s foundational settings.
9. I’ve created a kitchen sink template for now, to display the usage options for the Button component. This makes use of the get_component method found in our plugin. It works identical to the get_template_part, but allows us to pull in partials from our plugin. Setup a new page in Admin, and assign the template to be the Button Kitchen Sink. Once done, you should be able to visit the page, and see the 2 variants of the button output to the page. Provided you have the theme.json file setup correctly, you should see those styles inherited immediately. By removing the theme.json settings, you will see it fallback to the foundational styles within plugin. Currently, the plugin is enqueuing its foundational styles, but the idea is, we wouldn’t assume this, and would allow an Engineer to pick and choose what component styles they would like to be included in their final build. For now though, these have a very small footprint and I wanted to include these as a demo.
10. Obviously CSS Propery naming patterns will be key with this approach–definitely open for discussion here as well.
11. To make this partial available to Gutenberg, I’ve created a directory inside of the 10up-theme/includes/blocks directory called tui-button (again, naming up for grabs). The index.css file within this block can import the button style from the plugin directory directly. But for now, we have the plugin enqueuing it.
12. I also made updates to the main wp-content/package.json file to run the watches on the theme and plugins. This, in the scripts section. As well, I added the plugin to the workspaces entry in the same package.json file.
13. Within the 10up-theme, I added the tui-button entry under the 10up-toolkit in order to include it in the build.
14. Within the 10up-theme/includes/blocks directory I included the tui-button block folder. This is how the partial from our plugin becomes available to Gutenberg. Here you will see the markup.php file simply pulls in the component, and outputs it. For now. We may want to discuss how exactly this works. But this allows us to use the single source of truth. And it will receive an attributes array that should align with the args that appear as options in our actual block partial found in the plugin. The index.css file simply pulls in the button CSS file from the UI plugin as well.
15. Now I run an npm install in the root of the wp-content directory, then an npm run watch to watch the theme and the plugin. I added the hot flag to work with fast refresh.
16. I added a namepsace reference to the scaffold package.json file found in the theme root.
17. I added an import of the button styles to the frontend.js file in the theme for hot reloading
18. I added the SCRIPT_DEBUG constant to wp-config as true and also WP_DEBUG for good measure

**TODO:**
I would like to create several more variants of the Button. As well, the gutenberg implementation is still lack lustre, and needs to include the plugin styles. Currently, I'm simply outputting a Rich Text input to update the label on the button. I would like to spend time making this more cohesive with the front end.


themes/10up-theme/includes/blocks.php
```diff
@@ -69,6 +69,7 @@ function register_theme_blocks() {
 
 					// create helpful variables that will be accessible in markup.php file
 					$context = $block->context;
+					$wrapper_attributes = wp_kses_post( get_block_wrapper_attributes() );
```
fabiankaegy: @devinle We actually had this here previously but decided to move the `get_block_wrapper_attributes` into the actual markup.php files because you need to be able to pass properties to the `get_block_wrapper_attributes` function. 

Also it should not be escaped in that way. See #108

themes/10up-theme/includes/blocks/tui-button/edit.js
```diff
@@ -0,0 +1,42 @@
...
+
+	return (
+		<div {...blockProps}>
+			<RichText
+				tagName="h2"
+				placeholder={__('Learn more')}
+				value={label}
+				onChange={(label) => setAttributes({ label })}
+			/>
+			<ServerSideRender
```
fabiankaegy: I don't think that we should use this approach here. The editorial experience should be inline. If that means that for the beginning we need to maintain both the frontend and editor js in the scaffold UI plugin I think that is still worth it because of the better editorial experience.

devinle: Totally agree! I haven't worked on the editing experience. I've mostly done a test to see that the partial can be used the same way as we would in a template file. Editing experience needs to be cohesive. THANKS!

themes/10up-theme/includes/blocks.php
devinle: Done!

fabiankaegy:
Closing this for now

PR #115, opened 2022-05-02
Remove remnant quotation mark in attribute

kdo:
A tiny change, I've noticed an extra `"` in the block markup.


fabiankaegy (approved):
Great catch :)

PR #117, opened 2022-05-03
Update `theme.json` file to use version 2

fabiankaegy:
### Description of the Change

- Update the version of the `theme.json` file to 2.
- Update the name of the various options to match with how they are called in version 2.
- Removed the default palettes for Color, Gradient, and Duotone


joesnellpdx (approved):
@fabiankaegy approved

PR #121, opened 2022-05-10
Feature/tailwind v3 hybrid ian

iansvo:
### Description of the Change

This PR includes the following (see below). Additional details are provided for each.

Additional details on each are available below.

- Colors are now retrieved directly from theme.json
- Colors with the primary, secondary, and tertiary slugs are now safelisted via regex
- CSS organization has been reorganized around the use of imports, vs one tailwind index file
- Content definitions updated with list from Jon
- Tested/confirmed pattern for ensuring WP class defs are not lost

### Colors are now retrieved directly from theme.json
I grabbed the snippet I had created in my previous branch and brought it over here. One thing to note is that if we're using CSS variables to define colors we can't do opacity things like bg-primary/75 or something like that. Ultimately I think this is not an issue until we're told it is and it's a problem we don't need to be concerned with in our MVP. 

This pattern could easily be adapted to use any other data from theme.json. I think for now we should focus on keeping this simple and leaving it more open ended so people can determine how it needs to work for them and we can determine a better "one size fits all" solution out of the box.

### Colors with the primary, secondary, and tertiary slugs are now safelisted via regex
I did this slightly as a POC to just show what it looks like to do this. tailwind.config.js:65 has coverage for text and bg utilities for the primary, secondary, and tertiary colors. This ensures they don't get purged, even if they aren't used, which is helpful for when they're applied in other places, or something.

If a developer needs to add more to this, say some other palette color, they could do that pretty easily as needed by copying this snippet:

```
		{
			pattern: /(bg|text)-(bazinga)/,
		},
```

Where `bazinga` is the name of their custom color.

### CSS organization has been reorganized around the use of imports, vs one tailwind index file
The big reason I did this is to ensure that styles are being loaded in the correct order for the cascade. By default Tailwind includes styles in this order: base, component, utilities. The change I've made here ensures that all styles will cascade appropriately and reduce specificity issues with any custom base styles, custom abstractions, and utilities used as overrides.

### Content definitions updated with list from Jon
Our content globs now have the appropriate coverage without causing an infinite loop. My experience so far has been that we cannot use any kind of star (*) glob pattern at the root level or it will cause an infinite loop. This appears to be due to limitations around how webpack handles these watches, as noted in the Tailwind docs.

Ultimately this definition list should be suitable for 99% of use cases, developers will just need to add additional files located in the root directory to the definitions manually. 

### Tested/confirmed pattern for ensuring WP class defs are not lost
I've done some additional testing and research and I can confirm that any styles not applied inside of @layer rules (such as how wordpress.css is in this branch) those classes are retained even if they aren't detected in the source.

I think this moves the needle on several items for us. I know it's a lot so I appreciate you all taking the time to review. I'd love to discuss this in more detail if anyone has any thoughts or questions.


themes/10up-theme/assets/css/frontend/base/wordpress.css
```diff
@@ -50,3 +50,35 @@
...
+	* These will be applied to all headings with an `entry-title` class on the
+	* frontend and to the post heading in the block editor.
+	*
+	* The supplied styles are meant to match the default `h1` classes from
+	* Tailwind Typography.
+*/
+.edit-post-visual-editor__post-title-wrapper .editor-post-title__input,
+.entry-title {
+
+	@apply max-w-content mx-auto mb-6 text-6xl font-extrabold;
```
dainemawer, line 66: @Nikki-Jones @iansvo - do we want to be opinionated with these rules? Or should we just leave them untouched in terms of what `@tailwind/typography` provides?

themes/10up-theme/assets/css/frontend/style.css
```diff
@@ -4,28 +4,25 @@
...
+@import "tailwindcss/base";
 @import url("base/index.css");
 
-/* Layout - styles specific to layout */
 
-/* @import url("layout/index.css"); */
+/* Components */
+
+@import "tailwindcss/components";
+@import url("components/index.css");
```
dainemawer, line 16: @iansvo lets stick to the same standard here in this file - either lets use `url` or remove it like you have :)

themes/10up-theme/package.json
```diff
@@ -16,6 +16,9 @@
   },
   "devDependencies": {
     "@10up/stylelint-config": "^1.1.5",
+    "@tailwindcss/aspect-ratio": "^0.4.0",
```
dainemawer, line 19: @iansvo are we including these by default? It was my understanding that we were going to rather provide documentation around these than assume each project would want to have them implemented?

themes/10up-theme/tailwind.config.js
```diff
@@ -1,4 +1,28 @@
...
+const colors = require('tailwindcss/colors');
+const themeJSON = require('./theme.json');
+
+function getThemePalette() {
+	const { palette } = themeJSON.settings.color;
+
+	return Array.isArray(palette) ? palette : [];
+}
+
+function getThemeColors() {
```
dainemawer: Nice!

themes/10up-theme/tailwind.config.js
```diff
@@ -10,32 +34,60 @@ module.exports = {
...
+		{
+			pattern: /(bg|text)-(primary|secondary|tertiary)/,
+		},
 	],
 	theme: {
 		/* Override the default theme */
-
+		container: {
+			center: true,
+			padding: {
```
dainemawer, line 55: Do we want to override these values by default? It is quite an opinionated approach

themes/10up-theme/tailwind.config.js
```diff
@@ -10,32 +34,60 @@ module.exports = {
...
+				default: '1.25rem',
+				md: '2rem',
+				lg: '2.5rem',
+			},
+		},
 		/* Define brand colors using --wp custom properties from the theme.json */
 		colors: {
 			transparent: 'transparent',
 			current: 'currentColor',
-			white: 'var(--wp--preset--color--white)',
```
dainemawer, line 33: @iansvo will `--wp--preset--color--white` ever be anything other than `#fff` ?

themes/10up-theme/tailwind.config.js
```diff
@@ -10,32 +34,60 @@ module.exports = {
...
-			black: 'var(--wp--preset--color--black)',
+			...getThemeColors(),
+			/* Include additional palette colors as appropriate to fill in gaps for brand colors */
+			gray: colors.gray,
 		},
 		extend: {
 			/* Extend the default theme */
-
+			screens: {
+				'wp-md': '600px' /* WP Specific Breakpoint used in Gutenberg */,
```
dainemawer, line 72: Great add - would it be worth while aligning or adding specified wp media query values to this config in the long run?

themes/10up-theme/tailwind.config.js
fabiankaegy: This is really cool! :) 

One thing to note though is that this will not get you all colors that are used in a `theme.json` file. 

Every block can override the global color palette and therefore also add its own color palette... 

```jsonc
{
   "settings": {
        "color": {
            palette: [ ... ] // ← these are the global colors
        }
    },
    "blocks": {
        "core/button": {
             "color": {
                 "palette": [ ... ] // ← a color that does not exist on the global level can be introduced here
             }
        }
    }
}
```


And it doesn't account for the fact that `theme.json` in its nature is not singular but instead has different layers that have more / less specificity. 

The default setup looks like this: 




So the core / user-defined colors will not get picked up by the tailwind config. 

I'm not sure it really matters but just wanted to call it out.

themes/10up-theme/package.json
iansvo: Sorry that was leftover for when I was including them before. Removed in my latest commit.

themes/10up-theme/assets/css/frontend/style.css
iansvo: Doh! Copied from @layer and forgot to fix those. Updated in my latest commit.

themes/10up-theme/tailwind.config.js
iansvo: @dainemawer While that is the case, the value here is coming from a defined color (e.g. it doesn't just exist by default). You can see this in the Block Editor handbook example here: https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#css-custom-properties-presets-custom.

It is also worth noting that in general, true black/white isn't the color you'd want to use unless it's part of the brand guidelines. More often than not I'm seeing folks using offblack/white (think like `#fefefe` vs `#ffffff`, or `#0a0a0a` instead of `#000000`.

themes/10up-theme/tailwind.config.js
iansvo: I think we should absolutely include all core breakpoints if we can. This way we can do things like: 

`admin-md:hidden` or stuff like that. Our "toolbox" of styles should include anything we'd reasonably want to do on the frontend and ideally flow well with any block styles from core, so that seems like a natural next step.

themes/10up-theme/tailwind.config.js
iansvo: I'm sure everyone may have an opinion on this, but here's my thinking:

**Better DX**
After building 2-3 bigger projects, I think including the padding/centering like this as a default is way friendlier than having to add the centering/padding styles yourself. A couple of reasons for this:

It's way easier to just override the centering or padding then it is to add the equivalent styles each time.

Here's what it looks like to add the equivalent styles without that configuration:
```
<div class="container mx-auto sm:px-4 md:px-8 lg:px-10"> 
  ... 
</div>
```

Here's what it looks like to just completely remove the margin and padding while using the config I wrote:

```
<div class="container mx-0 px-0"> 
  ... 
</div>
```

Container styles come at the component layer so the utilities can easily override without having to account for every breakpoint that has padding by default. 

Also, the developer doesn't have to remember what container padding classes are needed by default, which helps promote consistency and reduce mental overhead (which are 2 of our larger objectives with the whole setup anyway). 

**Intuitive**
Virtually every single library/framework that uses a `container` abstraction has centering and padding. There are existing max-width utilities as well if someone just wants to have a particular maximum width at the upper end. So having this setup makes the container inherently more useful.

themes/10up-theme/tailwind.config.js
iansvo: @fabiankaegy That's a great point and funny enough I was literally just reading about this today, great minds think alike!

After giving it some thought it seems unlikely that we'd have a custom block color that isn't an existing color elsewhere. For instance, if we have a brand guide from a client its unlikely they have a single color that is only used on a single block or something. 

I feel like it wouldn't be a huge hassle to include those values in this setup. I'm pretty sure the config is smart enough to strip any duplicates (ex: there's a global color that is also defined in a block's colors) which further reduces the life there. 

As you said I'm also not sure it really matters, but just playing devil's advocate for a second...can you think of any scenarios where we'd define a color like that for a block that isn't in the theme color palette that we want to use in our code?

themes/10up-theme/assets/css/frontend/base/wordpress.css
Nikki-Jones: @iansvo @dainemawer I don't see the harm in being opinionated about this considering it will likely be overwritten and can be a starting point for anyone doing so

themes/10up-theme/tailwind.config.js
Nikki-Jones: @iansvo @dainemawer yeah I don't think we should get that opinionated with this. Although this is a nice way to cut down on some extra classes, it may not be always the approach we take to containers using responsive fixed widths

themes/10up-theme/assets/css/frontend/style.css
Nikki-Jones, line 16: @iansvo will we run in to any specificity issues by defining all components like this before the `@tailwind utilities` directive?

themes/10up-theme/tailwind.config.js
```diff
@@ -44,9 +96,10 @@ module.exports = {
 		},
 	},
 	plugins: [
-		/* Extend with 3rd party tailwind plugins or locally import your own */
-
-		/* Add Tailwind Prose https://tailwindcss.com/docs/typography-plugin */
-		tailwindProse(),
+		/*
+			Extend with 3rd party tailwind plugins or locally import your own
+			All plugins are included by default since they are not output if unused.
```
Nikki-Jones: @iansvo can we add a comment here explaining the use of prose in typographyPlugin?

themes/10up-theme/tailwind.config.js
```diff
@@ -44,9 +96,10 @@ module.exports = {
...
+		*/
+		typographyPlugin,
 	],
```
Nikki-Jones, line 105: @iansvo I see that you have added the other plugins to the package.json. Let's add them here too and just leave commented out

themes/10up-theme/tailwind.config.js
```diff
@@ -10,32 +34,60 @@ module.exports = {
...
 	safelist: [
-		/* Prevent editor-specific styles from being purged */
-		'editor-post-title__block',
-		'editor-post-title__input',
-		'entry-content',
-		'entry-title',
-		'block-editor-block-list__layout',
+		{
+			pattern: /(bg|text)-(primary|secondary|tertiary)/,
+		},
```
Nikki-Jones, line 49: @iansvo I think we could add this as a default:
`pattern: /(is|has|js)-*/,`

themes/10up-theme/assets/css/frontend/base/wordpress.css
dainemawer: Sounds good to me - lets make sure we note somewhere that that is our intention :) I will start putting a document together

themes/10up-theme/tailwind.config.js
dainemawer: Awesome, make it so!

themes/10up-theme/tailwind.config.js
dainemawer: I think we can add a section in the README for this as well with more detail :)

themes/10up-theme/tailwind.config.js
dainemawer: Sounds like a good approach :)

themes/10up-theme/tailwind.config.js
dainemawer: That makes sense @iansvo - and I agree, white and black do often get changed to slightly different tints. I think in this case, we should document this - the whole "teach a man to fish" speech. Any thoughts?

themes/10up-theme/tailwind.config.js
dainemawer: @iansvo you do make a good point. I like the idea of adding the `center` property. I think the `padding` / `margin` side of things is going to be very dependent on the project and the design. I've seen this change wildly on projects so maybe let's document this approach as it is a better practice overall and leave it up to the engineers to decide. We can certainly reimplement the `padding` / `margin` setup if we find that engineers ask for it. Great work and great forward thinking!

themes/10up-theme/tailwind.config.js
dainemawer: It may also be worth including support for `BEM` modifier classes here too!
`pattern: /(is|has|js)-|--*/,`

themes/10up-theme/tailwind.config.js
```diff
@@ -10,32 +34,60 @@ module.exports = {
...
+		'footer.php',
+		'front-page.php',
+		'functions.php',
+		'header.php',
+		'home.php',
+		'index.php',
+		'page.php',
+		'search.php',
+		'searchform.php',
+		'single.php',
```
dainemawer, line 35: Can we use a `glob` here to be able to support custom post types by default? Something like `single-*` and `archive-*`, `taxonomy-*` ?

themes/10up-theme/tailwind.config.js
fabiankaegy: @iansvo The example I was thinking about is a Call to action block for example that might want to expose 4 color variants: 

- `alert-red`
- `warning-orange`
- `success-green`
- `info-blue`

I would not expect these colors to be available in the global palette because you may not want to allow for text to be colored in red throughout the rest of the page. 

Again I don't think this is mandatory and such an edge case that we can worry about it later. But I think there is a relatively straightforward way to refactor this code to iterate over all `blocks['block-name'].settings.color.palette` values, dedupe them and then use that for the color config :)

themes/10up-theme/assets/css/frontend/base/wordpress.css
iansvo: @dainemawer I think Nikki is on the right track with providing a style for this in this manner. Overall I think we'd want to make sure that the value is correct for the normal typography sizing we use (e.g. if we're using prose-xl most of the time we'd probably want to use whatever that H1 font-size is) but we _do not_ want to just include `prose` classes all over. 

Those classes are not safe to drop on a big container that may include things that shouldn't receive those styles. The styles are written in such a way that it assumes you're placing them on the parent container of some long form content (think like blog post body, WYSIWYG editor output, that kinda stuff.

themes/10up-theme/assets/css/frontend/style.css
iansvo: @Nikki-Jones In terms of source order, utilities should always come last. Even in the default tailwind at-rules its listed the same way (base-->components-->utilities). So our own components should be loaded before these utilities. 

The idea would be that if we have any custom abstractions we've created (ex: buttons, menus maybe, that kind of thing) would want to be included prior to any of the actual utilities. This allows us to cleanly override things with the classes as expected and not create a specificity conflict where our abstraction is lower in the file than the utility and thus the utility no longer works in certain scenarios/breakpoints, etc. 

This pattern allows to leverage the same benefits of say, a single tailwind entry file with `@layer` blocks, while still having imported files and the like.

themes/10up-theme/tailwind.config.js
iansvo: I actually had these in from my previous changes, but I removed them in the latest push. 

I think if we're going to include them in package.json it makes sense to just have them loading all the time (except I guess forms). That said, I can't imagine a scenario in which we'd want to inherit form styles over the theme's own styles, but that might be fine for certain types of projects/situations. 

But things like line-clamp and aspect-ratio are so benign it feels like it might be confusing to _not_ include them, since so many examples on the web (Hyper, TailwindUI, and custom stuff) all make use of these. It seems like it would make the set up a bit more beginner friendly, as newbies might not realize these things are normally opt in. I of course don't want to include any unnecessary overhead on any project/tooling, but I think these are harmless and the benefits of having them built in seem worth it from what I can see. 

Does anyone have an actual reason why including these would be a negative, especially if we have documentation that adequately explains them and how they can be removed if you actually want to?

themes/10up-theme/tailwind.config.js
iansvo: To confirm, are these classes that we would be authoring in our own CSS somewhere? If so, they won't need to be safelisted unless they're included in an `@layer` which we don't need to use here anyway (at least not until its officially supported in browsers).

themes/10up-theme/tailwind.config.js
iansvo: @Nikki-Jones Are you thinking something like the comment that was there (sorry didn't quite mean to just nuke it) and then perhaps a link to the relevant portion of the README?

themes/10up-theme/tailwind.config.js
Firestorm980: @iansvo Do we want to include the "official" PurgeCSS safelist? https://purgecss.com/guides/wordpress.html

themes/10up-theme/assets/css/frontend/base/wordpress.css
dainemawer: Makes sense!

themes/10up-theme/tailwind.config.js
dainemawer: +1 on this just to be safe - no pun intended

themes/10up-theme/tailwind.config.js
dainemawer: I don't think there are any "negative" reasons not to include them. That being said, when it comes to extending core functionality - especially within the context of the `wp-scaffold` the approach has always been "opt-in" which to be honest, is kinda the opposite of what WordPress stipulates as a feature mantra but that's a totally different argument. 

One thing we need to keep in mind is the cognitive load of a feature like this, it takes a different approach in general to building sites, so our first priority should be to provide the option in a way that leaves the decision up to the discipline lead whether or not to use any functionality that is not considered critical. This may not apply to a simple line-clamp but in the past on projects, I've seen FEE, especially less senior engineers be thrown off and confused by far less complicated approaches than the above. 

We also, want to keep surprising the team right? So in my opinion, keep the plugins installed, comment them out like Nikki mentions and we can poll the FEE team to see if they would indeed use / like a feature like this. I have no doubt skilled FEE would consider it a no brainer (hence our debate) but we should always make sure to accommodate those who arent as technically astute as ourselves. 

</Rant>

@iansvo it may be an idea to poll the team now, out of the context of Tailwind to see who is even using `clamp` on projects - that way we can see if 70% of the team use it already, its a no brainer to include it. If only 10% of the team use it, well that means we need to educate, document and leave it up to them to decide if they want to implement in their projects :)

iansvo:
@dainemawer @Nikki-Jones @Firestorm980 

I have pushed some additional updates here that include the following:

1. I started up the README and included a few simple FAQ items so we can get the ball rolling on other contributions
2. All plugins are now included and only typography is actually loaded by default
3. My color importer now supports grabbing some or all of the block palette values, with built-in deduping to prevent overwrites. 

Let me know what you all think we need here before we can approve this and get things moved forward. I feel like we're in a really strong spot right now.

themes/10up-theme/tailwind.config.js
Nikki-Jones: @iansvo sounds great!

themes/10up-theme/assets/css/frontend/style.css
Nikki-Jones: @iansvo got it, thanks!

themes/10up-theme/includes/tailwind-helpers.js
```diff
@@ -0,0 +1,58 @@
+const themeJSON = require('../theme.json');
+
+function getThemePalette() {
```
Firestorm980, line 3: @iansvo Not huge, but let's make sure to add doc blocks to these functions to help explain what they are and why they're there.

themes/10up-theme/tailwind.config.js
```diff
@@ -1,4 +1,11 @@
-const tailwindProse = require('@tailwindcss/typography');
+const colors = require('tailwindcss/colors');
+
+// Tailwind Plugins
+const typographyPlugin = require('@tailwindcss/typography');
+const { getThemeColors } = require('./includes/tailwind-helpers');
+// const formsPlugin = require('@tailwindcss/forms');
+// const aspectRatioPlugin = require('@tailwindcss/aspect-ratio');
+// const lineClampPlugin = require('@tailwindcss/line-clamp');
```
Firestorm980, line 8: @iansvo @Nikki-Jones @dainemawer Do we want to leave these comments here so people can easily bring them back or should we just leave them out entirely and explain it in the docs?

themes/10up-theme/tailwind.config.js
Firestorm980: I'm also for documenting more than setting it by default. Just because not all our projects do this because of other factors. I like the idea of having that information _somewhere_ though so that it can be used / modified if desired.

Firestorm980 (commented):
Some comments. Nothing crazy. Well done!

themes/10up-theme/includes/tailwind-helpers.js
iansvo: @Firestorm980 I was just thinking about that. I think if we're settled on the functionality my next step can be adding the doc blocks all around and adding support documentation in the readme, all in all will take perhaps 15-30 minutes tops.

themes/10up-theme/tailwind.config.js
iansvo: Unfortunately, that is not possible due to the issues with webpack's glob handling that I outlined previously in that word doc you made for us. See here for additional details: https://v2.tailwindcss.com/docs/just-in-time-mode#styles-rebuild-in-an-infinite-loop

In short, if we do this in the root directory it causes an infinite loop (which is the original real I had the problem and @Nikki-Jones didn't, since she was targeting subfolders with her globs). 

Until we can find a way to address that or otherwise not use Webpack, I think that's the best we'll be able to do without a pretty big lift and definitely what we'd want to do for the MVP.

iansvo: We can't currently do this due to the way webpack handles globs, which I had noted before in the document you started originally. See here for details: https://tailwindcss.com/docs/content-configuration#styles-rebuild-in-an-infinite-loop

For our MVP, I'd say we need to consider this a documentation/training point and leave it at that. The only way around this is to fundamentally address the way the globs are handled or otherwise not use webpack. Neither of these things are small LOE or straightforward, and the effort for the dev to just do it and have it referenced in the README seems much smaller by comparison.

themes/10up-theme/tailwind.config.js
iansvo: My thought was to include them by default and let someone decide if they want to remove them. This way we err on the side of simplicity and straightforwardness and if someone is bothered they can always remove at their discretion.

themes/10up-theme/tailwind.config.js
dainemawer: Awesome, lets err on the side of simplicity for the Beta. Thanks @iansvo !

themes/10up-theme/includes/tailwind-helpers.js
dainemawer: DocBlocks would be a huge help, regarding the README I'd like to see a lot of collaboration on that doc - its going to be really important to have that tied up and making sense to the team!

themes/10up-theme/tailwind.config.js
dainemawer: Hmm, I'm undecided on this one. The only negative thing about having them in the code base is that it kinda looks like a half-baked feature, so maybe the README is better. Considering we still need to document these plugins regardless, maybe the README is better.

PR #122, opened 2022-05-11
fix traverse blocks in theme.json to get colors that aren’t used in global palette

fabiankaegy:
### Description of the Change

Ensure any colors that may only be used in a block color palette in the `theme.json` also get added to the tailwind color config. See https://github.com/10up/wp-scaffold/pull/121#discussion_r868958442


fabiankaegy:
Choosing as this was already accomplished in [f752cc5](https://github.com/10up/wp-scaffold/pull/121/commits/f752cc5e5e1e659a59ec69ec52ea060fd92f1d70)

PR #124, opened 2022-05-18
Add extensions.json in .vscode config

dainemawer:
This PR adds an extension recommendation to the Tailwind project.
You can read more about [Extension Recommendations on the VSCode site](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions).

This update recommends that the following extension be used on all Tailwind projects: https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss


iansvo (approved):
Looks good, I just want to confirm if we need to test or if you were able to validate locally.

dainemawer:
@iansvo gave this a test now. There are a couple of things to note.

The `.vscode` folder must be in the root of the workspace in order for VS CODE to "suggest" the extension. 
I've placed `.vscode` in the repo (so it can be tracked) - if a user does not open `wp-content` as the root of the project/workspace then the recommendation notification will not trigger. The screenshot below will only appear the very first time a user opens a workspace.

Screenshot 2022-05-24 at 07 04 59

The same directory predicament applies when VS CODE displays the declared extension in Extensions > Recommended, however, this panel in VS Code persists regardless of whether its the first time you've opened up the workspace.

Screenshot 2022-05-24 at 07 14 14

As far as I can tell we either need to advise users that they must open `wp-content` as the root of their project, or we'll need to alter the 10updocker script to generate a `.vscode` folder when it spins up a new site.

Let me know if you have any thoughts.

cc @Nikki-Jones @Firestorm980

iansvo:
@dainemawer Thanks for the details here, this is great. I think we should stick with our previous principle of optimizing for the majority, in which case I'd say we should keep your pattern and include the file in the repo.

As far as I'm aware, the WordPress implementation of 10up-toolkit is explicitly designed to be used in this way, so anything outside of that would be either a non-WordPress or non-standard usage. So I think we can just include it here and that should do it.

PR #132, opened 2022-06-08
Adding unmissable notice (error) when using HMR

Antonio-Laguna:
### Description of the Change

It's been a common issue that when using [@10up/toolkit ](https://github.com/10up/10up-toolkit/tree/develop/packages/toolkit) the `--hot` flag was being used but `SCRIPT_DEBUG` wasn't being set to `true` on `wp-config.php`. This has been causing some frustration since there's no error. Nothing shows up on the terminal or the browser's console that might indicate you're missing anything. 

This adds a check using `wp_get_environment_type` (check [docs](https://developer.wordpress.org/reference/functions/wp_get_environment_type/)) which safely defaults to `production`. Both 10up_docker and wp_env are setting this correctly. If the check passes then it checks whether the `fast-refresh` file exists and if it does but `SCRIPT_DEBUG` isn't `true` then it dies showing this notification:

image

Even if a bit drastic, this is a controlled scenario in which a developer is running `--hot`. This should point developers to the appropriate place to look for documentation and get hot module replacement working again.

### Checklist:

- [X] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.

### Changelog Entry

* Changed current HMR file detection so it only happens whenever `wp_get_environment_type` is either `development` or `local`. If the file is detected and `SCRIPT_DEBUG` isn't set to `true`, `wp_die` is invoked pointing the user (a developer) to the appropriate resource to achieve the intended outcome.

### Credits

Props @fabiankaegy for brainstorming, nudging and providing copy


themes/10up-theme/functions.php
```diff
@@ -14,12 +14,32 @@
 define( 'TENUP_THEME_INC', TENUP_THEME_PATH . 'includes/' );
 define( 'TENUP_THEME_BLOCK_DIR', TENUP_THEME_INC . 'blocks/' );
 
+$check_hmr =               wp_get_environment_type() === 'local' || 'development';
```
nicholasio:
```suggestion
$check_hmr =               in_array( wp_get_environment_type(), [ 'local', 'development' ], true );
```

unless I'm missing anything, the current code is going to default to 'development' which is going to be truthy on production envs.

Antonio-Laguna: Good catch!

nicholasio (approved):

fabiankaegy:
@Antonio-Laguna Can you post an update here as to the current state of things? Are we ready to merge this? Is this dependent on a toolkit update?

Antonio-Laguna:
@fabiankaegy I think this should be good to go but I'd rely on @nicholasio. I think the file generated by Toolkit is the one that handles it now and I see we're pointing to _almost_ the last version. I'm not sure though with `wp_get_environment_type` now.

Antonio-Laguna:
@fabiankaegy & @nicholasio this has been updated :)

fabiankaegy:
@Antonio-Laguna can you make sure to also update the plug file?

Antonio-Laguna:
Done! Thanks for bringing that up!

fabiankaegy (approved):
This looks good to me 🚀 @nicholasio can I get a final 👍 from you before merging?

nicholasio (approved):
LGTM

Issue #138, opened 2022-06-13
Include 10up-toolkit in NPM dependencies instead of devDependencies


johnwatkins0:
### Is your enhancement related to a problem? Please describe.

I suggest reconsidering this project's approach to dev dependencies vs. dependencies.

Any production build/deployment process for a project based on WP Scaffold should have a mechanism to exclude dependencies that are not needed within the automated job -- specifically, we should be able to run `npm install --production` in a build script to exclude `devDependencies`, similarly to how we use `composer install --no-dev`. This can have a noticeable impact on build times wherever there are multiple dev dependencies.

I suggest differentiating dev dependencies vs. production dependencies this way: 

- A dev dependency is a dependency that's **not needed to build, run, or deploy a production site**. husky and lint-staged would fall in this category because they're local development tools. A testing library like jest would also fall in this category because, even though it's used in automated test jobs, it's not needed in the final build/deployment script that builds the production version of a project.
- A production dependency is a dependency **used to build, run or deploy the production version of a project**. This includes 10up-toolkit, as 10up-toolkit is crucial to build any project that uses the theme or plugin from WP Scaffold.

So my suggestion is:

- Keep husky and lint-staged where they are, because they're not needed for anything outside local development.
- Move 10up-toolkit into dependencies

This would allow engineers to selectively run `npm install --production` in build/deploy scripts wherever it makes sense to do so.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #139, opened 2022-06-14
Workspaces friendlier packages

Antonio-Laguna:
### Description of the Change

The main purpose of this change was to make sure that `package-lock` files can't be created within internal packages. Related is some package.json tweak to avoid publishing packages by mistake and also some housekeeping around the README file.

### Checklist:

- [X] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This scaffold is the starting point for all 10up WordPress projects.
 
-It contains a bare bones theme and must use plugin for you to base your development off of. Asset bundling is handled entirely by [10up Toolkit](https://github.com/10up/10up-toolkit).
+It contains a bare-bones theme and must use plugin for you to base your development off of. Asset bundling is handled entirely by [10up Toolkit](https://github.com/10up/10up-toolkit).
 
 ## Requirements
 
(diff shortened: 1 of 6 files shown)


mu-plugins/10up-plugin/.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
+package-lock=false
```
nicholasio, line 2: does this prevent any actual issues? What happens if someone run npm install inside the theme or plugin? Wouldn't this just make it harder to detect someone mistakenly running npm install inside themes/plugins?

README.md
```diff
@@ -25,15 +25,15 @@ It contains a bare bones theme and must use plugin for you to base your developm
...
 
 ```json
 	"watch:theme": "npm run watch -w=tenup-theme",
 	"watch:plugin": "npm run watch -w=tenup-plugin",
 	"watch": "run-s watch:theme watch:plugin",
 ```
-7. To add npm dependencies to your theme and/or plugins add the `-w=package-name` flag to the `npm install` command. E.g: `npm install --save prop-types -w=tenup-plugin` **DO NOT RUN** `npm install` inside an individual workspace/package. Always run the from the root folder.
-8. If you're building Gutenberg blocks and importing `@wordpress/*` packages, **you do not** need to manually install them as `10up-toolkit` will handle these packages properly.
+7. To add npm dependencies to your theme and/or plugins add the `-w=package-name` flag to the `npm install` command. E.g: `npm install --save prop-types -w=tenup-plugin`. **DO NOT RUN** `npm install` inside an individual workspace/package. Always run it from the root folder.
+8. If you're building Gutenberg blocks and importing `@wordpress/*` packages, **you do not** need to manually install them as `10up-toolkit` will handle these packages properly. However, it might still be a good idea for intellisense and code completion.
```
nicholasio, line 36: I'm not sure how I feel about this: `However, it might still be a good idea for intellisense and code completion.` I believe @johnwatkins0 used to this but then gave up? It would be nice if we had a types only package IMO cc @fabiankaegy 

At the very least this would have to be devDeps.

johnwatkins0: Yup, I personally like having those packages installed locally for code completion in VS Code, but I have also found that other engineers are confused to see them in package.json and this can lead to some incorrect assumptions when attempting to debug editor issues, which can waste time (e.g. folks will think they need to update the `@wordpress/*` dependencies , when these packages in fact aren't bundled into the code). If these packages _are_ installed, though, I agree they belong in devDependencies as they're for local dev only (see my related suggestion at https://github.com/10up/wp-scaffold/issues/138)

Antonio-Laguna: The point of the change was to not sound as harsh and rather give the option. You don’t need to but it’s not harmful. Given there's no mention of whether this needs to be a `devDependencies` or not I didn't think it was needed to add this but can definitely do

mu-plugins/10up-plugin/.npmrc
Antonio-Laguna: So there's a non-obvious behavior. 

If this isn't added, the `package-lock` can be created at the package level rather than the root level. From my testing, this happens on npm 7 but not on npm 8 which, no matter where you install from, always goes to the root.

Now, if you go and install without any package-lock from a package with this setup, the root gets a package-lock.json instead of the package getting one. I'd rather not advertise that it's OK to install dependencies from packages with these safeguards and stick to the idea that ONLY installing from the root is the way.

Can't foresee any downside to this change. Is there anything I'm missing? Would you rather wait till NPM 8 is normalized?

fabiankaegy:
@Antonio-Laguna Whats the state of this PR? :)

Antonio-Laguna:
@fabiankaegy this is up to whoever can merge, I think there's value in this PR but not strictly needed. Was there something you wanted to get out?

fabiankaegy:
@Antonio-Laguna I was trying to move through some of the open PR's in this repo to hopefully reduce the number of open PR's... =D And based on the various conversations in this one I wasn't sure whether it would be ready to be merged or closed or whether updates would be needed :)

fabiankaegy (approved):
From my perspective and my testing the code here is valuable and ready to be merged. 

@nicholasio does that work for you or do you have objections (See https://github.com/10up/wp-scaffold/pull/139/files#r897212574)

darylldoyle (approved):
Much like @fabiankaegy, if there are no specific problems that this is going to cause, I see no reason not to merge this PR.

nicholasio:
recent versions of npm will actually default to npm install from root even if you run directly from a workspace so I'm not sure if this is needed.

fabiankaegy:
> recent versions of npm will actually default to npm install from root even if you run directly from a workspace so I'm not sure if this is needed.

Seems it's no longer an issue. Closing for now. We can always reopen

PR #143, opened 2022-06-24
add scaffold for link and button style to help engineers with editor style bleed

fabiankaegy:
### Description of the Change

Add scaffold for link and button base CSS in order to help engineers work around introducing bleed of editor styles into the WordPress Admin interface. 

Since we have switched to using the [`add_editor_style`](https://developer.wordpress.org/reference/functions/add_editor_style/) function for including our editor styles in #69 we've found two common issues where base styles of link and button elements affect some instances of Block Editor components. This of course is not intended behavior and really should be fixed in WordPress core. See WordPress/gutenberg#10178 

To work around this I've gone ahead and added an empty shell for the link and button base styles so that they exclude the WordPress core elements. In order to not increase the specificity of the selector, we are using the `:where` pseudo-selector. 

Closes #

### Alternate Designs

We in theory introduce a separate frontend-only stylesheet again to split the styles that should get applied in the editor apart. We've seen in the past however, that this quickly leads to many inconsistencies and editorial experiences that don't match the frontend appearance. 

### Possible Drawbacks

The `:where` pseudo selector is currently supported in 90% of browsers according to [caniuse.com](https://caniuse.com/?search=%3Awhere). This should cover all the browsers we are targeting for client builds.

### Verification Process

- Add some custom styles to either the `link.css` files in the `assets/css/frontend/base` directory
- Navigate to the editor
- Insert a paragraph and add a link. -> See that the styles are showing up correctly 

- Insert an Image Block -> See on the Placeholder state that the link in the placeholder that is part of core does not get affected. 

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

Fixed: Remove editor-style of links and buttons bleeding into Core WordPress elements

### DIFF
--- a/themes/10up-theme/assets/css/frontend/base/button.css
+++ b/themes/10up-theme/assets/css/frontend/base/button.css
@@ -2,7 +2,7 @@
  * Ensure that we are not overriding the editor styles of the
  * WordPress Components.
  *
- * This is a workarround for: https://github.com/WordPress/gutenberg/issues/10178
+ * This is a workaround for: https://github.com/WordPress/gutenberg/issues/10178
  *
  * using :where to prevent the specificity increase of using :not
  */
--- a/themes/10up-theme/assets/css/frontend/base/button.css
+++ b/themes/10up-theme/assets/css/frontend/base/button.css
@@ -1,0 +1,11 @@
+/*
+ * Ensure that we are not overriding the editor styles of the
+ * WordPress Components.
+ *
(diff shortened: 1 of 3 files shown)


Antonio-Laguna (approved):
This looks good to me and it's also a great enhancement!

joesnellpdx (approved):
also approved by me! this is great!

PR #151, opened 2022-07-24
Add PurgeCSS WordPress SafeList

dainemawer:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Changed - Existing functionality
> Deprecated - Soon-to-be removed feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @username, @username2, ...

### Checklist:

### DIFF
--- a/themes/10up-theme/tailwind.config.js
+++ b/themes/10up-theme/tailwind.config.js
@@ -45,8 +45,8 @@ module.exports = {
 		'templates/**/*.php',
 	],
 	safelist: [
+		...purgecssWordpress.safelist,
 		{
-			...purgecssWordpress.safelist,
 			pattern: /(bg|text)-(primary|secondary|tertiary)/,
 		},
 	],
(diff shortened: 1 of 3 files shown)


themes/10up-theme/tailwind.config.js
```diff
@@ -45,6 +46,7 @@ module.exports = {
 	],
 	safelist: [
 		{
+			...purgecssWordpress.safelist,
 			pattern: /(bg|text)-(primary|secondary|tertiary)/,
 		},
 	],
```
Firestorm980: @dainemawer @iansvo Are we sure this is in the right spot? https://tailwindcss.com/docs/content-configuration#using-regular-expressions Based on the docs, I get the feeling it is more supposed to be like:
```suggestion
	safelist: [
		...purgecssWordpress.safelist,
		{
			pattern: /(bg|text)-(primary|secondary|tertiary)/,
		},
	],
```
I could be wrong though and it may not matter. FWIW, in my testing it appeared to work the same. Just want to make sure we're following the guidance.

iansvo: Yeah that is actually how it should be. The object there is only necessary because we're using a pattern. I believe this safelist here is just an array of strings, so I would say lets change it as Jon suggests.

dainemawer: Great catch! Thanks @iansvo @Firestorm980  !

PR #155, opened 2022-08-01
Example Block: Remove tabs

roseg43:
Removed cases where tabs are used instead of spaces for indenting, so that the spacing of the file is consistent.

### Description of the Change
The `block.json` file for the example block inconsistently uses tabs and spaces for indentation. I've fixed any cases where tabs are being used, as the majority of the document was using spaces before this fix.

This fixes an issue where indenting would appear inconsistent on platforms like GitLab / GitHub when viewing assets.

### How to test the Change
Open `themes/10up-theme/includes/blocks/example-block/block.json` in your editor of choice and confirm that all indentation uses a consistent character.

### Changelog Entry

> Fixed - Bug fix
Example Block: Fix instances where tabs were being mixed with spaces for indentation in `block.json`

### Credits

Props @roseg43 (wasn't sure if I should add myself, so here I am 😂)

### Checklist:

### DIFF
--- a/themes/10up-theme/includes/blocks/example-block/block.json
+++ b/themes/10up-theme/includes/blocks/example-block/block.json
@@ -1,16 +1,16 @@
 {
   "$schema": "https://schemas.wp.org/trunk/block.json",
   "apiVersion": 2,
-	"title": "Example Block",
-	"description": "An Example Block",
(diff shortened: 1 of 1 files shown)


fabiankaegy (approved):
Thanks for the fix :)

PR #158, opened 2022-08-12
Feature/auto initialize modules

darylldoyle:
Currently there is very little PHP scaffolding in the repo, especially in the MU-Plugin. This means that to set up, engineers usually end up copy/pasting the scaffold from other projects. This PR proposes a starting point for a less obstructive PHP scaffold, where PHP classes are automatically registered based on conditionals.

### Description of the Change

This PR does the following:

* Adds a new abstract `Module` class that defines `can_register()` and `register()` methods.
* Utilizes the [haydenpierce/class-finder](https://packagist.org/packages/haydenpierce/class-finder) package to find all classes within a namespace (currently `TenUpPlugin`)
* Of all the found classes, find those that extend the `Module` class.
* Of those classes, find those where the `can_register()` method returns `true`.
* Calls the `register()` method on all those remaining.

There is some documentation available within the `docs` directory, see [here](https://github.com/10up/wp-scaffold/blob/feature/auto-initialize-modules/docs/registering-classes.md).

This is related to #4

### How to test the Change
Clone the repo into a WordPress install and try registering a class using the new method.

### Changelog Entry

> Added - Auto-registration of PHP classes in the MU-plugin

### Credits
Props @darylldoyle 

### Checklist:

### DIFF
--- a/docs/registering-classes.md
+++ b/docs/registering-classes.md
@@ -1,12 +1,12 @@
 # Registering Classes in the MU-Plugin
 
-The MU-Plugin utilizes a system to uniformly, auto-register classes that lie within its namespace. Whilst there are a few constraints, it eases the requirements for engineers to add their classes to multiple locations each time they add one to the system.
+The MU-Plugin and the theme utilize a system to uniformly, auto-register classes that lie within their namespaces. Whilst there are a few constraints, it eases the requirements for engineers to add their classes to multiple locations each time they add one to the system.
 
 To do this, it uses the [haydenpierce/class-finder](https://packagist.org/packages/haydenpierce/class-finder) package, which reads the `composer.json` file to help locate files that belong in certain namespaces.
 
 ## How do I define a class to be auto-registered?
 
-All you need to do to get a class to auto-register is extend the `TenUpPlugin\Module::class` class. That will require you to implement a `can_register()` and a `register()` method.
+All you need to do to get a class to auto-register is extend the `TenUpPlugin\Module::class` or `TenUpTheme\Module::class` classes. That will require you to implement a `can_register()` and a `register()` method.
 
 ### `can_register()`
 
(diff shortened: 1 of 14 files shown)


dsawardekar:
@darylldoyle I think this is a step in the right direction. It would make modules easily manageable like single file mu-plugins. One thing that stood out to me was the composer.json requirement. On most projects it is not an issue to set this as a deploy step. 

I would suggest adding an admin_notice if composer is not found in case it hasn't been setup or failed to load etc.

docs/registering-classes.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+## How do I get an instance of my registered class?
+
+The old way of doing this would be to use the `get_plugin_support()` function. As we no longer define and register our classes in the same way, this doesn't work.
+
+The best way now, is to use the `get_class()` method that's part of the `ModuleInitialization` class.
+
+```php
+$site_settings = \TenUpPlugin\ModuleInitialization::instance()->get_class( '\TenUpPlugin\Admin\SiteSettings' );
```
dsawardekar: @darylldoyle I would suggest adding a Helper like `TenUpPlugin\get_module` to make this shorter.

mu-plugins/10up-plugin/includes/classes/ModuleInitialization.php
```diff
@@ -0,0 +1,140 @@
...
+		return false;
+	}
+
+	/**
+	 * Get all the initialized classes.
+	 *
+	 * @return array
+	 */
+	public function get_all_classes() {
+		return $this->classes;
```
dsawardekar, line 155: @darylldoyle I would suggest adding a filter hook here in case some customization is needed, like init order etc.

mu-plugins/10up-plugin/includes/core.php
```diff
@@ -22,7 +23,7 @@ function setup() {
 	};
 
 	add_action( 'init', $n( 'i18n' ) );
-	add_action( 'init', $n( 'init' ) );
+	add_action( 'init', $n( 'init' ), 8 );
```
dsawardekar: @darylldoyle I would add a filter hook to allow control over the `init` priority.

dsawardekar (changes requested):

mu-plugins/10up-plugin/includes/classes/ModuleInitialization.php
darylldoyle: This is just used to return a list of all initialized classes, I don't foresee it being used often, so it was just included as an edge case. 

That said, I have added a `load_order` property to the `Module` class that can be overwritten. This works much like WordPress priorities, as in it allows us to choose which classes load first!

docs/registering-classes.md
darylldoyle: Good call

mu-plugins/10up-plugin/includes/core.php
darylldoyle: Makes sense!

darylldoyle:
> @darylldoyle I think this is a step in the right direction. It would make modules easily manageable like single file mu-plugins. One thing that stood out to me was the composer.json requirement. On most projects it is not an issue to set this as a deploy step.
> 
> I would suggest adding an admin_notice if composer is not found in case it hasn't been setup or failed to load etc.

Thanks for the notes @dsawardekar! I've updated the PR to fix those and also to add the loader to the theme as suggested by @gthayer.

If either of you have another chance for a review, that'd be much appreciated!

dsawardekar (approved):
@darylldoyle This looks good to me. It will add much needed standardization to the MU plugin. And it is also pretty lightweight and does not require a big ramp up.

tlovett1:
Hey all, I'm realizing 0 documentation was added for this. Can someone please do that ASAP? @darylldoyle @dsawardekar

darylldoyle:
@tlovett1 there is some documentation available in https://github.com/10up/wp-scaffold/blob/trunk/docs/registering-classes.md.

I can also write up a blog post for the internal blog if that'd help?

Is there anything additional you'd like to see added to the documentation?

PR #160, opened 2022-08-16
Feature/19138441 tailwind v3 testing

Nikki-Jones:
### Description of the Change

After using Tailwind with 10up Toolkit on a real project there were some changes made to the base configuration. These modification help Wordpress Gutenberg styling from conflicting with Tailwind's reset. The developer experience is also improved with the addition of Stylelint and nesting.

### Changelog Entry

> Added - Stylelint and stylelint config
> Changed - Postcss config
> Changed - Tailwind preflight reset to custom reset
> Changed - style.css import structure

### Checklist:


themes/10up-theme/.stylelintrc
```diff
@@ -0,0 +1,17 @@
+{
+	"extends": [
+		"@10up/stylelint-config"
+	],
+	"rules": {
+		"at-rule-no-unknown": [
+			true,
+			{
+				"ignoreAtRules": [
```
dainemawer, line 9: Great addition!!

themes/10up-theme/assets/css/frontend/tailwind/index.css
```diff
@@ -0,0 +1,47 @@
+/* stylelint-disable at-rule-no-unknown */
```
dainemawer: This is a far better approach - nice work - do we still need the above stylelint comment if we've amended the stylelint config?

themes/10up-theme/assets/css/frontend/style.css
```diff
@@ -4,25 +4,33 @@
  *
  */
 
+/* Tailwind */
```
dainemawer: An overall question for you @Nikki-Jones @iansvo - does `@apply` support nesting? 

```
.some-css-rule {
  @apply p-4;

  & .some-child-rule {
   @apply m-3;
  }
}
```

dainemawer (approved):
@Nikki-Jones some great additions here, really stoked! 👍🏻 from me!

themes/10up-theme/assets/css/frontend/style.css
iansvo: @dainemawer I can confirm this functions out of the box: 

image

themes/10up-theme/.stylelintrc
```diff
@@ -0,0 +1,17 @@
...
+					"screen",
+					"layer",
+					"tailwind"
```
iansvo, line 12: @Nikki-Jones Would `@apply` already be considered known, or does that need to be added?

themes/10up-theme/TAILWIND.md
```diff
@@ -44,8 +44,7 @@ Tailwind allows you to author custom classes that group multiple utilities toget
 		border-blue-800
 		hover:bg-blue-800
 		hover:text-white
-		rounded-lg
-		bg-white;
```
iansvo, line 48: Why do we need to remove the background here?

themes/10up-theme/assets/css/frontend/base/wordpress.css
```diff
@@ -51,7 +51,9 @@
 	@apply max-w-none;
 }
 
-/**
+@layer components {
```
iansvo: @Nikki-Jones Is the intent here to make these styles below purgeable? Or are they supposed to always be in the CSS output?

Things like `.editor-post-title__input` look like a Block Editor class, which wouldn't be picked up in our content patterns since it's not in the actual theme.

themes/10up-theme/assets/css/frontend/style.css
```diff
@@ -4,25 +4,33 @@
...
+@import url("tailwind/index.css");
```
iansvo, line 33: @Nikki-Jones From what I can see it looks like the idea here is that you want the entire tailwind file included separately so everything else comes after those styles, unless someone chooses to use the `@layer` rule? 

This is actually how we had this initially, but at a point I made an adjustment to what is in the epic now for a few reasons, but the biggest one is it somewhat limits our flexibility with the cascade in a way that the default scaffold does not (since everything is included in a certain order in the scaffold out of the box). 

For instance, if I wanted to always include a class in our bundle, but insert it at the same point as other "component styles" (which allows the Tailwind utilities to take priority still and cleanly override when selector weight is equal) I wouldn't be able to do that because anything I leave out of an `@layer` rule would come after everything else (including tailwind's built-in utilities). This type of thing can cause a utility to not be able to override a component style (picture a button where you need to change 1-2 small things about it and nothing else, so you add 2 utilities with your button class) because the utilities are no longer as specific as they are by default. 

I've only built around 3 sites with WordPress and Tailwind so far so I couldn't say exactly how often this might happen, but it strikes me as a likely possibility we can avoid by just organizing things differently (like we are now). This allows the developer to either include their code at the correct point and always have it be present, or opt-in to using `@layer` based on the situation. 

I'd love to get everyone's take on this and see what we all thing. cc: @Firestorm980 @dainemawer

themes/10up-theme/assets/css/frontend/tailwind/reset.css
```diff
@@ -0,0 +1,241 @@
+/*
+Custom Reset modified for WordPress based off of Tailwind's reset:
```
iansvo: @Nikki-Jones Can you elaborate a bit on what changes you made to be more suitable for WordPress? I haven't really encountered this before but I'm interested to understand.

themes/10up-theme/postcss.config.js
```diff
@@ -1,8 +1,27 @@
...
+/* eslint-disable global-require */
+
+module.exports = () => {
+	const config = {
+		plugins: [],
+	};
+
+	// Only load postcss-editor-styles plugin when we're processing the editor-style.css file.
+
+	config.plugins.push(require('postcss-import'));
```
iansvo: @Nikki-Jones Why are we using push here? I'm not sure what value the extra function call gives us here when 3/4 are one line. Plus it's a pretty well established convention to just define in the array literal, especially if we're not conditionally adding anything to the array.

themes/10up-theme/tailwind.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
 	 * Any missing sections will fall back to Tailwind’s default configuration.
 	 * https://github.com/tailwindlabs/tailwindcss/blob/master/stubs/defaultConfig.stub.js
 	 */
+	corePlugins: {
+		/**
+		 * Preflight injects this file https://github.com/tailwindlabs/tailwindcss/blob/master/src/css/preflight.css which
```
iansvo, line 21: Solid catch here, this is great!

themes/10up-theme/tailwind.config.js
```diff
@@ -81,6 +93,7 @@ module.exports = {
 		/*
 			Add prose classes to provide sensible typography styles to longform content blocks
 			See: https://tailwindcss.com/docs/plugins#typography
+			Comment out this plugin to add project specific typography styles
```
iansvo: @Nikki-Jones I feel like this comment is kind of misleading. It reads as though you either have to use this plugin, or custom typography styles. 

The main thing this plugin does is just add the `.prose-*` classes, which have to be specifically applied to be used. The other typography configuration is handled entirely in the config, so I feel like someone who doesn't know Tailwind would be very easily confused by this. 

```suggestion
			You may remove this plugin if you're not using the `.prose-*` classes for long-form content.
```

themes/10up-theme/assets/css/frontend/tailwind/index.css
iansvo: I imagine we wouldn't need both, but I think the discrete file is more scalable for sure.

iansvo (changes requested):
@Nikki-Jones I'm glad you were able to get this put together and there's some great stuff in here to help tighten up the DX/ergonomics. 

I have 1-2 things that I would like to get the team's take on before we merge this and make sure we're all on the same page, and I believe there's perhaps a few small cleanup items as well.

I tried to leave some detailed feedback just let me know if you (or anyone else) want to talk through anything.

themes/10up-theme/TAILWIND.md
Nikki-Jones: @iansvo it was repeated in the lines above :)

themes/10up-theme/.stylelintrc
Nikki-Jones: @iansvo yes @apply is known/not tailwind specific

themes/10up-theme/assets/css/frontend/tailwind/reset.css
Nikki-Jones: @iansvo the explanation is here - https://github.com/10up/wp-scaffold/blob/3eb3750d7c3bd8944c500f8022f4b712660ab66a/themes/10up-theme/tailwind.config.js, but let me know if you think it's worth adding in both spots

themes/10up-theme/assets/css/frontend/style.css
Nikki-Jones: @iansvo as per our slack discussion, take a look with how I've structured things now, let me know what you think!

dainemawer: I really like the details here @Nikki-Jones ! Very helpful - @iansvo - if you're happy will you get this merged?

iansvo: I feel better about this with the cascade for sure. Thanks so much for talking through it with me. 

I would not that I'm not sure we really need to expose an option to change where variants go. But I don't want to just nit pick endlessly. We can iterate on a few things like that once this thing gets in front of people and see if it actually confuses/footguns anyone or not. 

I really appreciate you taking the time to comment these things and explain more of the mentality behind it. I think as we're looking to increase adoption that will be helpful to make this more friendly to new devs. 

Good stuff!

themes/10up-theme/assets/css/frontend/base/wordpress.css
```diff
@@ -76,9 +77,12 @@
 	* classes, please see:
 	*
 	* https://tailwindcss.com/docs/typography-plugin
+	*
+	* To disable the richtext styling remove the `prose` class.
 */
 .block-editor-block-list__layout,
 .entry-content {
```
iansvo, line 84: I'm not sure we want to associate prose implicitly with `entry-content`. While it makes sense from a convention standpoint (entry content shouldn't be used to wrap complex layouts) so I think we can let it ride for now, but part of me thinks its better/easier (and more the "Tailwind way") to not do this, since we control all instances of where `entry-content` appears (vs say, the `block-editor-block-list__layout` class which comes from Gutenberg). 

@dainemawer @Firestorm980 What do you two think on this? I think we could possibly let it go in our MVP and just see if we/anyone dislikes it later. I just wanted to make sure we discussed it.

themes/10up-theme/assets/css/frontend/tailwind/reset.css
```diff
@@ -0,0 +1,246 @@
...
+	h2,
+	h3,
+	h4,
+	h5,
+	h6 {
+		font-size: inherit;
+		font-weight: inherit;
+	}
+
+	a:where(:not(.components-external-link)) {
```
iansvo, line 60: Eyyyy that `:where`. Nice!

iansvo (commented):
Lookin pretty nice! I just had one point of discussion I wanted to get the group's take on but otherwise I'm with it!

I really love the attention to detail on this CSS reset (the default normalize is...not great even in the best of scenarios). 

Really nice work here.

themes/10up-theme/assets/css/frontend/base/wordpress.css
dainemawer: Yeah, I think on this one let's see what happens when engineers pull this down and begin work. This will definitely require some tweaking here and there! Resolving this comment for now, but thanks for bringing it up @iansvo

dainemawer (approved):

themes/10up-theme/assets/css/frontend/tailwind/components.css
```diff
@@ -0,0 +1,16 @@
...
+* Traditionally these would be classes like card, btn, link etc.
+* These styles will be imported at the @import url("tailwindcss/components") level
+* in the main stylesheet
+*
+* More info: https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes
+*/
+
+@layer components {
+
+}
```
Firestorm980, line 16: @iansvo @Nikki-Jones I don't know if we've tried this so let me know if I'm crazy, have we combined `@layer` and `@import`? I realize why we'd want a layer, but I'm just thinking about if a engineer wants a lot of their own components.

Firestorm980 (approved):

themes/10up-theme/assets/css/frontend/tailwind/components.css
Nikki-Jones: Hey @Firestorm980, could you add an example?

Firestorm980: @Nikki-Jones So, for example, let's say I have 10+ custom components. Would I do:
```css
@layer components {
  @import url('components/card');
  @import url('components/select');
  /* ... a bunch of other imports */
}
```
Or, something else? (It occurs to me you may `@import` and then use `@layer` in the component itself.)

And to be clear, I'm not saying something is wrong with this code, more an implementation question that you may or may not have the answer to.

themes/10up-theme/assets/css/frontend/base/wordpress.css
Nikki-Jones: @iansvo @dainemawer I'm a 100% for removing prose if that's what you're suggesting. I think it may cause more headaches and confusion than it's worth

iansvo (approved):

PR #168, opened 2022-12-01
updates theme.json to v2 for compatibility with filtering

jacobdubail:
### Description of the Change

This updates the theme.json version to v2 for compatibility with the new theme.json filters. It also updates keys to match v2 schema.

### How to test the Change

Confirm custom styles are applying to the theme.

### Checklist:

### DIFF
--- a/themes/10up-theme/theme.json
+++ b/themes/10up-theme/theme.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://schemas.wp.org/wp/6.0/theme.json",
+  "$schema": "https://schemas.wp.org/wp/6.1/theme.json",
   "version": 2,
   "settings": {
     "color": {
--- a/themes/10up-theme/theme.json
+++ b/themes/10up-theme/theme.json
@@ -1,5 +1,6 @@
 {
-  "version": 1,
+  "$schema": "https://schemas.wp.org/wp/6.0/theme.json",
+  "version": 2,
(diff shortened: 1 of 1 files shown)


themes/10up-theme/theme.json
```diff
@@ -1,5 +1,6 @@
 {
-  "version": 1,
+  "$schema": "https://schemas.wp.org/wp/6.0/theme.json",
```
fabiankaegy: We should use 6.1 at this point :)

fabiankaegy (approved):
Thanks for the update! Added one comment that I think should be changed before merging but otherwise good to go 👍

themes/10up-theme/theme.json
jacobdubailcadmv: Thanks for catching this! I pushed a new commit that updates the schema to 6.1

fabiankaegy (approved):

fabiankaegy:
@jacobdubail If you want to you can also change the base of this PR to the `trunk` branch. We will want to use the update for the entire scaffold :)

joesnellpdx:
@jacobdubail 

We want to have the filter for the theme.json updates in here -right? Or is that still up for discussion?

joesnellpdx:
ohh... @jacobdubail I see how you are doing it now.. and that's actually pretty slick!!

PR #170, opened 2023-01-06
auto versioning for editor styles

hugosolar:
### Description of the Change

We found some instances where we included editorStyles but these are enqueued using the WordPress Core version by default and this can be problematic each time we update those styles since the file are being long-term cached by the environment. I made this fix to ensure we're automatically registering the styles with a dynamic version and then just passing the style handler to the block.json.

### How to test the Change
Run `npm run install && npm run build` and then test the theme in a WordPress install.
check the network panel once you enter to the block editor and look for the CSS file included in the editor as part of the example-block. The file should have the `ver` parameter something like `?ver=1672948490` instead of `?ver=6.1.1`

### Credits
Props @hugosolar 

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,5 +17,10 @@
   "scripts": {
     "lint": "phpcs .",
     "lint-fix": "phpcbf ."
+  },
+  "config": {
+    "allow-plugins": {
+      "dealerdirect/phpcodesniffer-composer-installer": true
+    }
   }
 }
--- a/themes/10up-theme/includes/blocks.php
+++ b/themes/10up-theme/includes/blocks.php
@@ -61,8 +61,21 @@ function register_theme_blocks() {
 		foreach ( $block_json_files as $filename ) {
 
 			$block_folder = dirname( $filename );
+			$block_name   = basename( $block_folder );
 
 			$block_options = [];
+
+			/**
+			 * Register block editor style with automated versioning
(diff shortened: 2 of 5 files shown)


fabiankaegy:
Thanks for creating this PR. I will have to spend a little more time with it early next week to see whether there are alternate approaches. 

I was unaware of the different handling for styles compared to scripts. I had assumed it also uses the generated version from the `asset.php` file. But that isn't the case. Instead core expects the `block.json` file to contain a `version` which will get used as the cash invalidator. If no version is set the core version will get used. 

> The current version number of the block, such as 1.0 or 1.0.3. It's similar to how plugins are versioned. This field might be used with block assets to control cache invalidation, and when the block author omits it, then the installed version of WordPress is used instead.

See: https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/block-api/block-metadata.md#version

This here is where it actually happens in core: https://github.com/WordPress/wordpress-develop/blob/5a497225b286964ff9302a21272731172840ed05/src/wp-includes/blocks.php#L249

hugosolar:
@fabiankaegy I was thinking about having `{file}.asset.php` for those styles but not sure how to handle that with `10up-toolkit`. Do you think that could be a better approach?

fabiankaegy:
@hugosolar Thanks again for your suggestions in here. Since this has been resolved in toolkit version 5.1 we can close this PR here.

PR #172, opened 2023-02-15
Fix issue with static methods in class autoloader

darylldoyle:
### Description of the Change
In #158 we added a way to auto-initialise and load classes in the MU plugin and theme.
Because the classfinder uses static methods, it only works on a single application root.

To get around this, we need to manually set the application root before we look for classes in a namespace, so that it can find the right composer.json file

This has been tested on a client project that's using this version of the scaffold.

### How to test the Change
- Add a class to the MU plugin and another to the theme. 
- Make sure that both are registered.

### Changelog Entry
* Fixed - Issue with class loading conflict.

### Checklist:

### DIFF
--- a/mu-plugins/10up-plugin/includes/classes/ModuleInitialization.php
+++ b/mu-plugins/10up-plugin/includes/classes/ModuleInitialization.php
@@ -56,7 +56,9 @@ class ModuleInitialization {
 	 * @return array
 	 */
 	protected function get_classes() {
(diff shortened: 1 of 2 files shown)


darylldoyle:
@joesnellpdx @tlovett1 wondering if one of you could review this since it fixes a bug that could cause engineers some pain in the scaffold.

christianc1 (approved):

fabiankaegy (approved):
Tested this on a project and it works great (y)

PR #173, opened 2023-03-09
Fix incorrect formatting in all editorconfig files

bjorn2404:
### Description of the Change

Fixes formatting error for `crlf` line in all repository .editorconfig files. This may have been originally copied over from [wordpress-deveop](https://github.com/WordPress/wordpress-develop/blob/master/.editorconfig) (see link for correct usage). I initially thought of just removing `wp-config-sample.php` from the files since the purpose isn't 100% clear but instead updated the PR to the correct format.

image

* Multiple items need to be wrapped with curly braces. Docs on [Glob Expressions](https://spec.editorconfig.org/#glob-expressions).

Closes #

### Credits

Props @bjorn2404 

### Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,5 +11,5 @@ indent_style = tab
 indent_style = space
 indent_size = 2
 
-[*.txt]
+[{*.txt,wp-config-sample.php}]
 end_of_line = crlf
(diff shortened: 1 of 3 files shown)


darylldoyle (approved):
Thanks @bjorn2404, this is great!

PR #176, opened 2023-05-03
Fix mu-plugin entry file referencing wrong variable

theskinnyghost:
### Description of the Change
The mu-plugin main file was referencing a PHP constant defined in the theme, which gets loaded after the mu-plugin causing the site to break.

### How to test the Change
Install WP scaffold and setup composer/npm. The site should just work.

### Credits
Props @theskinnyghost 

### Checklist:


fabiankaegy (approved):
Thanks for the fix :)

PR #184, opened 2023-05-24
Remove manifest.json

### LINKED ISSUE #118: Rethink inclusion of `manifest.json` inclusion in scaffold
One thing I've noticed again and again over the past few years is that the default values in the `manifest.json` file don't get changed. Most sites don't require it and don't get built as PWA's at this point. And since more browsers started using the `theme_color` for styling bits of their UI this can lead to unfortunate visuals. 

https://github.com/10up/wp-scaffold/blob/ce3177347c810bd0cba5f18e9de45afba499eef9/themes/10up-theme/manifest.json#L9

| iOS | macOS |
|-----|--------|
|Simulator Screen Shot - iPhone 12 - 2022-05-03 at 08 23 49|CleanShot 2022-05-03 at 08 27 26@2x|

If used correctly the `theme_color` is very useful. But currently, I think that it more often gets "forgotten" and stays with the default value which is creating a bad experience.

tlovett1:
Removes `manifest.json` and the code adding it to `wp_head`. We don't need to make our projects PWAs. Also, no teams are actually updating the manifest so our client websites have app install buttons that say "10up Theme".

Screen Shot 2023-05-24 at 9 31 11 AM


fabiankaegy:
Thank you for doing this. I raised that a while ago in #118 but didn’t get much feedback. 

I’m all on board with this

tlovett1:
@fabiankaegy wish I'd seen that. Sorry!

PR #185, opened 2023-06-06
Fix with a problem related to composer install, added .prettierignore & npm update

bmarshall511:
### Description of the Change

4th day here at 10up and working through the orientation for engineering in setting up the wp-scaffold. Ran into an issue when first setting it up related to composer. Specifically when running `composer install`. Got the following error:

```
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
Your lock file does not contain a compatible set of packages. Please run composer update.
```

image

To fix, updated composer & included additional info in the README.md should devs run into this in the future.

Also, added a .prettierignore for those using prettier to format code, ignoring markdown files so avoid it removing spaces that are intended.

### How to test the Change
On a fresh setup, run `composer install` in the root of `wp-content`.

### Credits
Props @bmarshall511 

### Checklist:
- [ x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [x ] I have updated the documentation accordingly.


.prettierignore
darylldoyle, line 1: Thanks, this should be useful! Can we add a trailing line break to this file, please?

README.md
```diff
@@ -24,25 +24,38 @@ It contains a bare bones theme and must use plugin for you to base your developm
...
+
+ When running `composer install`, you may encounter the following warning:
+
+ ```bash
+ Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended  that you run `composer update` or `composer update <package name>`.
+ Your lock file does not contain a compatible set of packages. Please run composer update.
+ ```
+
+ To resolve this issue, simply execute the `composer update` command.
+
```
darylldoyle, line 38: **question:** Do we still need this addition now that you've updated the lock file?

package-lock.json
darylldoyle, line 1: **issue:** Can we remove this file, please? We should leave it up to the projects to set the lock file, as per https://github.com/10up/wp-scaffold/pull/41

darylldoyle (changes requested):
Thanks for the contribution @bmarshall511 and it's great to have you here. I have a couple of thoughts below, when you have a moment!

.prettierignore
bmarshall511: Definitely! I'll get that added.

README.md
bmarshall511: Absolutely! 😄 I totally considered that too! But I thought it'd be helpful to keep that addition for future reference, especially as new updates roll in. It can be a handy tip for fellow newbies like me. However, I do agree that the error itself is pretty self-explanatory, so it might be redundant to include it in the README. Let me know what you think.

package-lock.json
bmarshall511: Absolutely! I'll get rid of it. I'll remember not to include them in the future.

.gitignore
```diff
@@ -7,6 +7,7 @@ phpunit.xml
 uploads
 upgrade
 cache
+package-lock.json
```
fabiankaegy: Hey @bmarshall511 👋 

(Thank you for diving in head first and directly starting to contribute back to our tooling :))

This here is an annoying piece thing about the setup we currently have with the scaffold... We want the packages to be on the latest version when the project is initiated. Therefore we don't want to include the lock file here in the `wp-scaffold` template repository. 

But we do want every project to use lock files.

Because of that we cannot add the `package-lock.json` file to the `.gitignore` here. 

I would love to find a better way to make it easier for maintainers to automatically not include the lock file but still ensuring that actual projects then use the lock file. So if you have any ideas we are always open to new ways of doing this :) 

For now please remove this addition here.

bmarshall511: Completely understand, I've got that updated so it won't be ignored.

PR #190, opened 2023-07-31
fix: prettier and eslint

nicholasio:
### Description of the Change

~`@wordpress/eslint-plugin` has been an optional peer deep for a while and needs to be installed at the project level.~ UPDATE: It is installed automatically by latest version of toolkit.

This PR pins `prettier` to a specific (and supported version). Prettier will be something projects will need to chose the version as updating prettier in toolkit has been an issue and locking down prettier to a specific version has also been an issue so in the next major of toolkit it will only declare a version range.

### Changelog Entry

> Added - New feature
> Changed - Existing functionality
> Deprecated - Soon-to-be removed feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @username, @username2, ...

### Checklist:

### DIFF
--- a/mu-plugins/10up-plugin/package.json
+++ b/mu-plugins/10up-plugin/package.json
@@ -15,7 +15,7 @@
     "node": ">=12.0.0"
   },
   "devDependencies": {
-    "10up-toolkit": "^5.2.1"
+    "10up-toolkit": "^5.2.2"
   },
   "dependencies": {
     "prop-types": "^15.7.2"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,18942 @@
+{
+  "name": "tenup-wp-scaffold",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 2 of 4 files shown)


JakePT:
~Worked for me.~

Sorry, spoke too soon. It was only working when there wasn’t an `.eslintrc` file in the project root. I’d removed it while testing other things and when I restored it ESLint stopped working with this error:

```
Error: Failed to load config “@10up/eslint-config/wordpress” to extend from.
```

This PR seems to cause 10up-toolkit to be installed in the workspace’s own `node_modules` directory, instead of the project root, which might be responsible.

If include an `.eslintrc` file Visual Studio Code and the `10up-toolkit lint-js` commands continue to work as expected, but the pre-commit hooks don’t, because they’re using the root eslintrc file, and it causes this error.

nicholasio:
@JakePT would you mind trying again?

JakePT:
@nicholasio Having prettier installed in wp-content (at 2.6.2 or 2.8.8 works for me) is working for me with toolkit 5.2.2.

PR #193, opened 2023-08-23
rename core block overrides JS file to block editor script for greater clarity

fabiankaegy:
### Description of the Change

The `core-block-overrides` file is named incorrectly. In reality this file is used for all kinds of things in the block editor. Because of the name though, folks are often confused as to where other types of customization such as new meta fields etc should be placed.

### DIFF
--- a/themes/10up-theme/includes/core.php
+++ b/themes/10up-theme/includes/core.php
@@ -189,7 +189,7 @@ function enqueue_block_editor_scripts() {
 	wp_enqueue_script(
 		'block-editor-script',
 		TENUP_THEME_DIST_URL . 'js/block-editor-script.js',
-		UTility\get_asset_info( 'block-editor-script', 'dependencies' ),
+		Utility\get_asset_info( 'block-editor-script', 'dependencies' ),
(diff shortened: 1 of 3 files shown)


themes/10up-theme/includes/core.php
```diff
@@ -185,18 +185,14 @@ function admin_scripts() {
...
-			$dep['dependencies'],
-			$dep['version'],
-			true
-		);
-	}
+function enqueue_block_editor_scripts() {
+	wp_enqueue_script(
+		'block-editor-script',
+		TENUP_THEME_DIST_URL . 'js/block-editor-script.js',
+		UTility\get_asset_info( 'block-editor-script', 'dependencies' ),
```
Antonio-Laguna:
```suggestion
		Utility\get_asset_info( 'block-editor-script', 'dependencies' ),
```

Antonio-Laguna (commented):
@fabiankaegy just a minor typo and then it's good to go!

Antonio-Laguna (approved):

PR #195, opened 2023-09-08
Add scaffolding command for quickly generating blocks

fabiankaegy:
### Description of the Change

This PR adds a new `scaffold:block` command to the 10up starter theme which allows anyone to easily scaffold out a new block. It is based on the `@wordpress/create-block` package with custom block templates that are bundled in the scaffold itself. 

Closes #

### How to test the Change

To test this you can run `npm run scaffold:block -w themes/10up-theme` in the `wp-content` folder. This will run through a few prompts and generate the block for you. 

### Changelog Entry

> Added - Block Scaffolding command to easily build out new blocks

### Credits

Props @fabiankaegy 

### Checklist:

### DIFF
--- a/bin/create-block-template/index.js
+++ b/bin/create-block-template/index.js
@@ -28,6 +28,9 @@ module.exports = {
 	variants: {
 		default: {},
 		innerBlocks: {},
+		withViewScript: {
+			viewScript: 'file:./view.js',
+		},
 	},
 	blockTemplatesPath: join(__dirname, 'block-templates'),
 };
--- a/bin/create-block-template/index.js
+++ b/bin/create-block-template/index.js
@@ -1,0 +1,33 @@
+/**
+ * External dependencies
+ */
+const { join } = require('path');
+
+module.exports = {
+	defaultValues: {
+		slug: 'example-block',
+		category: 'text',
+		title: 'Example Block',
+		description: 'Example Block',
+		attributes: {},
+		supports: {
+			html: false,
+		},
+		customBlockJSON: {
+			textdomain: 'tenup',
+		},
+		namespace: 'tenup',
+		wpScripts: false,
+		wpEnv: false,
+		version: false,
+		folderName: './src/blocks/example-block',
+		render: 'file:./markup.php',
+		editorStyle: false,
+		style: 'file:./style.css',
+	},
+	variants: {
+		default: {},
+		innerBlocks: {},
+	},
+	blockTemplatesPath: join(__dirname, 'block-templates'),
+};
--- a/bin/create-block-template/block-templates/edit.js.mustache
+++ b/bin/create-block-template/block-templates/edit.js.mustache
@@ -1,0 +1,21 @@
+{{#isDefaultVariant}}
+import { useBlockProps } from '@wordpress/block-editor';
+{{/isDefaultVariant}}
+{{#isInnerBlocksVariant}}
+import { useBlockProps, useInnerBlocksProps } from '@wordpress/block-editor';
+{{/isInnerBlocksVariant}}
+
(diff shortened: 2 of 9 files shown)


fabiankaegy:
@nicholasio I'd love to get your take on this. Do you think this should live in the scaffold here or in toolkit. 

My reason for including it in the scaffold is that I want project teams to be able to modify the starter block templates to their linings. Which would be more difficult in toolkit since it doesn't get exposed. 

However if it were in toolkit it would be easier to ship updates to the scaffolding tool compared to here where essentially it gets stale once the project is scaffolded unless someone manually updates it.

bin/create-block-template/index.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * External dependencies
+ */
+const { join } = require('path');
+
+module.exports = {
+	defaultValues: {
+		slug: 'example-block',
```
darylldoyle, line 8: **question (non-blocking):** Should this by dynamically set by the script?

bin/create-block-template/index.js
```diff
@@ -0,0 +1,36 @@
...
+		category: 'text',
+		title: 'Example Block',
+		description: 'Example Block',
```
darylldoyle, line 11: question (non-blocking): Should these by dynamically set by the script?

bin/create-block-template/index.js
```diff
@@ -0,0 +1,36 @@
...
+		attributes: {},
+		supports: {
+			html: false,
+		},
+		customBlockJSON: {
```
darylldoyle, line 16: **question (non-blocking):** Maybe this is ignorance on my part, but what does this property do?

darylldoyle (commented):
I love this! Some way to do this has always been at the top of my mind. 

I've left a couple of questions, but once those are answered it looks good!

ivanlopez:
@fabiankaegy If we were to add it to the toolkit, we could also create a command to publish the templates into the project to be customized, like Laravel does here: https://laravel.com/docs/10.x/artisan#stub-customization.

bin/create-block-template/index.js
fabiankaegy: Yeah :) This value here only sets the default if the user doesn't provide one in the CLI prompts.

(See attached Video)


https://github.com/10up/wp-scaffold/assets/20684594/fe0562ed-c37b-417d-a077-bd3183f298c1

bin/create-block-template/index.js
fabiankaegy: Same answer as here: https://github.com/10up/wp-scaffold/pull/195#discussion_r1321655475

bin/create-block-template/index.js
fabiankaegy: https://github.com/WordPress/gutenberg/blob/trunk/packages/create-block/docs/external-template.md#defaultvalues

> customBlockJSON (no default) - allows definition of additional properties for the generated block.json file.

This allows us to add any other additional properties to the `block.json` files generated :)

fabiankaegy:
@ivanlopez 👀 That does sound like an interesting solution that would cover both cases!

nicholasio:
@fabiankaegy Since the actual logic to create scaffold the block is handled by `@wordpress/create-block` I don't think there's much value in moving this to toolkit as it would essentially wrap `@wordpress/create-block`.  One thing that toolkit could do it let projects override the default template, so if nothing is provided at the project level it would use the one shipped with toolkit.

With that said, it's probably fine to ship this directly in the scaffold for now.

fabiankaegy:
Okay, in that case, let's keep it simple for now. 

We can ship this as is and iterate on it if we find we want more flexibility in the future

PR #216, opened 2024-02-21
Update phpcs.xml exclude pattern for plugins folder

### LINKED ISSUE #198: PHPCS does not scan mu-plugins folder
https://github.com/10up/wp-scaffold/blob/trunk/phpcs.xml#L5

This pattern should be ./plugins/*

Jamesking56:
### Description of the Change

Updates excluded directory for plugins to not also capture `mu-plugins`.

Closes #198

### How to test the Change

Test if PHPCS scans `mu-plugins`

### Changelog Entry

> Fixed - Bug fix
Fixed `mu-plugins` directory not being scanned by PHPCS

### Credits

@Jamesking56

### Checklist:

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -2,7 +2,7 @@
 <ruleset name="10up PHPCS">
 	<description>10up PHPCS extended.</description>
 	<exclude-pattern>dist/</exclude-pattern>
-	<exclude-pattern>plugins/*</exclude-pattern>
+	<exclude-pattern>./plugins/*</exclude-pattern>
 	<rule ref="10up-Default" />
 
 	<arg name="colors"/>


phpcs.xml
```diff
@@ -2,7 +2,7 @@
 <ruleset name="10up PHPCS">
 	<description>10up PHPCS extended.</description>
 	<exclude-pattern>dist/</exclude-pattern>
-	<exclude-pattern>plugins/*</exclude-pattern>
+	<exclude-pattern>./plugins/*</exclude-pattern>
```
fabiankaegy, line 5: Is it possible to add an override for a exclude? 

In that case I would like to say somthing like this: (pseudo code)

```xml
    <exclude-pattern>./plugins/*</exclude-pattern>
    <exclude-pattern>./mu-plugins/*</exclude-pattern>
    <include-pattern>./mu-plugins/tenup-plugin</include-pattern>
```

So that any other installed 3rd party MU plugins don't get checked. 

Essentially the goal should be to only check first party code 🤔 I would kind of want this to just use the `.gitignore` rules.

darylldoyle: Just noting that `<include-pattern>` can only be used inside specific rules. We should use the `<file>` directive to add specific directories, though it seems that `<exclude-pattern>` will override those added via `<file>`.

It may be that we need to use some convoluted regex to allow specific directories if we blanket ignore `./plugins/*`

PR #218, opened 2024-03-14
Feature/php post type and taxonomy scaffold

### LINKED ISSUE #175: Add PHP Scaffold for CPTs and Taxonomies

All new 10up projects are based on this repo, but there is very little PHP scaffolding in place. This means that each different project ends up with a slightly different implementation for post types, taxonomies, REST endpoints etc.

This has two effects:

* Slows down the initial setup of a project, as the tech lead has to put these scaffolds in place.
* Increases the cognitive load for engineers that are switching between projects, as they have to learn the differences in setups.

### Designs

Since the auto-initialisation of modules was merged in #158, it's allowed us to create much more modular scaffolds, with no requirement for factory classes etc.

I propose introducing abstract classes that contain all the base information needed to register a CPT/Taxonomy, which can then be extended to implement project-specific ones.

For example, the following abstract post type would then allow us to easily implement further ones:

```php
<?php
/**
 * AbstractPostType
 *
 * @package TenUpPlugin
 */

namespace TenUpPlugin\PostTypes;

use TenUpPlugin\Module;

/**
 * Abstract class for post types.
 */
abstract class AbstractPostType extends Module {

	/**
	 * Get the post type name.
	 *
	 * @return string
	 */
	abstract public function get_name();

	/**
	 * Get the singular post type label.
	 *
	 * @return string
	 */
	abstract public function get_singular_label();

	/**
	 * Get the plural post type label.
	 *
	 * @return string
	 */
	abstract public function get_plural_label();

	/**
	 * Default post type supported feature names.
	 *
	 * @return array
	 */
	public function get_editor_supports() {
		$supports = [
			'title',
			'editor',
			'author',
			'thumbnail',
			'excerpt',
			'revisions',
		];

		return $supports;
	}

	/**
	 * Get the options for the post type.
	 *
	 * @return array
	 */
	public function get_options() {
		$options = [
			'labels'            => $this->get_labels(),
			'public'            => true,
			'has_archive'       => true,
			'show_ui'           => true,
			'show_in_menu'      => true,
			'show_in_nav_menus' => false,
			'show_in_rest'      => true,
			'supports'          => $this->get_editor_supports(),
		];

		return $options;
	}

	/**
	 * Get the labels for the post type.
	 *
	 * @return array
	 */
	public function get_labels() {
		$plural_label   = $this->get_plural_label();
		$singular_label = $this->get_singular_label();

		// phpcs:disable -- ignoring template strings without translators placeholder since this is dynamic
		$labels = array(
			'name'                     => $plural_label,
			// Already translated via get_plural_label().
			'singular_name'            => $singular_label,
			// Already translated via get_singular_label().
			'add_new_item'             => sprintf( __( 'Add New %s', 'tenup-plugin' ), $singular_label ),
			'edit_item'                => sprintf( __( 'Edit %s', 'tenup-plugin' ), $singular_label ),
			'new_item'                 => sprintf( __( 'New %s', 'tenup-plugin' ), $singular_label ),
			'view_item'                => sprintf( __( 'View %s', 'tenup-plugin' ), $singular_label ),
			'view_items'               => sprintf( __( 'View %s', 'tenup-plugin' ), $plural_label ),
			'search_items'             => sprintf( __( 'Search %s', 'tenup-plugin' ), $plural_label ),
			'not_found'                => sprintf( __( 'No %s found.', 'tenup-plugin' ), strtolower( $plural_label ) ),
			'not_found_in_trash'       => sprintf( __( 'No %s found in Trash.', 'tenup-plugin' ), strtolower( $plural_label ) ),
			'parent_item_colon'        => sprintf( __( 'Parent %s:', 'tenup-plugin' ), $plural_label ),
			'all_items'                => sprintf( __( 'All %s', 'tenup-plugin' ), $plural_label ),
			'archives'                 => sprintf( __( '%s Archives', 'tenup-plugin' ), $singular_label ),
			'attributes'               => sprintf( __( '%s Attributes', 'tenup-plugin' ), $singular_label ),
			'insert_into_item'         => sprintf( __( 'Insert into %s', 'tenup-plugin' ), strtolower( $singular_label ) ),
			'uploaded_to_this_item'    => sprintf( __( 'Uploaded to this %s', 'tenup-plugin' ), strtolower( $singular_label ) ),
			'filter_items_list'        => sprintf( __( 'Filter %s list', 'tenup-plugin' ), strtolower( $plural_label ) ),
			'items_list_navigation'    => sprintf( __( '%s list navigation', 'tenup-plugin' ), $plural_label ),
			'items_list'               => sprintf( __( '%s list', 'tenup-plugin' ), $plural_label ),
			'item_published'           => sprintf( __( '%s published.', 'tenup-plugin' ), $singular_label ),
			'item_published_privately' => sprintf( __( '%s published privately.', 'tenup-plugin' ), $singular_label ),
			'item_reverted_to_draft'   => sprintf( __( '%s reverted to draft.', 'tenup-plugin' ), $singular_label ),
			'item_scheduled'           => sprintf( __( '%s scheduled.', 'tenup-plugin' ), $singular_label ),
			'item_updated'             => sprintf( __( '%s updated.', 'tenup-plugin' ), $singular_label ),
			'menu_name'                => $plural_label,
			'name_admin_bar'           => $singular_label,
		);
		// phpcs:enable

		return $labels;
	}

	/**
	 * Registers a post type and associates its taxonomies.
	 *
	 * @uses $this->get_name() to get the post's type name.
	 * @return Bool Whether this theme has supports for this post type.
	 */
	public function register() {
		$this->register_post_type();
		$this->register_taxonomies();

		$this->after_register();

		return true;
	}

	/**
	 * Registers the current post type with WordPress.
	 *
	 * @return void
	 */
	public function register_post_type() {
		register_post_type(
			$this->get_name(),
			$this->get_options()
		);
	}

	/**
	 * Registers the taxonomies declared with the current post type.
	 *
	 * @return void
	 */
	public function register_taxonomies() {
		$taxonomies = $this->get_supported_taxonomies();

		$object_type = $this->get_name();

		if ( ! empty( $taxonomies ) ) {
			foreach ( $taxonomies as $taxonomy ) {
				register_taxonomy_for_object_type(
					$taxonomy,
					$object_type
				);
			}
		}
	}

	/**
	 * Returns the default supported taxonomies. The subclass should declare the
	 * Taxonomies that it supports here if required.
	 *
	 * @return array
	 */
	public function get_supported_taxonomies() {
		return [];
	}

	/**
	 * Run any code after the post type has been registered.
	 *
	 * @return void
	 */
	public function after_register() {
		// Do nothing.
	}

}
```

To implement a project-specific CPT, we'd then do something like:

```php
<?php
/**
 * Demo Post Type
 *
 * @package TenUpPlugin
 */

namespace TenUpPlugin\PostTypes;

/**
 * Demo post type.
 */
class Demo extends AbstractPostType {

	/**
	 * Get the post type name.
	 *
	 * @return string
	 */
	public function get_name() {
		return 'tenup-demo';
	}

	/**
	 * Get the singular post type label.
	 *
	 * @return string
	 */
	public function get_singular_label() {
		return esc_html__( 'Demo', 'tenup-plugin' );
	}

	/**
	 * Get the plural post type label.
	 *
	 * @return string
	 */
	public function get_plural_label() {
		return esc_html__( 'Demos', 'tenup-plugin' );
	}

	/**
	 * Can the class be registered?
	 *
	 * @return bool
	 */
	public function can_register() {
		return true;
	}

	/**
	 * Supported post type features.
	 *
	 * @return array
	 */
	public function get_editor_supports() {
		return [
			'title',
			'editor',
			'thumbnail',
			'page-attributes',
			'excerpt',
			'revisions',
			'custom-fields',
		];
	}

	/**
	 * Get the options for the post type.
	 *
	 * @return array
	 */
	public function get_options() {
		return array_merge(
			parent::get_options(),
			[
				'hierarchical' => false,
				'rewrite'      => [
					'slug' => 'demo',
				],
			]
		);
	}

	/**
	 * Returns the default supported taxonomies. The subclass should declare the
	 * Taxonomies that it supports here if required.
	 *
	 * @return array
	 */
	public function get_supported_taxonomies() {
		return [
			'tenup-tax-demo',
		];
	}
}
```

Due to the fact that we have auto-initializing modules, the engineer would only need to create the sub-class and fill it out. There's no need to add anything to a factory, meaning less merge conflicts.

We can also do something very similar for taxonomies.

This solution would provide engineers with:

* A standardised, extendable way to register post types
* A pre-existing scaffold within the repo
* Loss cognitive overhead when switching between projects.

It also gives us a nice lead-in to creating a tool to help scaffold these items quickly, much like has been proposed in #89

Darshan has also submitted https://github.com/10up/wp-scaffold/issues/104. I don't _feel_ like this is an alternative to that approach, but it could be complimentary to it by proving a start for that modular scaffold.

### Code of Conduct

darylldoyle:
### Description of the Change
This adds a PHP Scaffold for adding Post Types and Taxonomies to the scaffold. 

Whilst this isn't _needed_, it means that implementation would be very similar across projects, reducing cognitive load when context switching. It would also allow engineers to pick up newer projects much quicker.

Below is a screencast of me using this to create a new post type and taxonomy.
https://capture.dropbox.com/L7fP0jGYEvLobArN

Closes 10up/wp-scaffold#175 

### How to test the Change
Check out the branch and try it out

### Credits

Props @darylldoyle 

### Checklist:

### DIFF
--- a/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
+++ b/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
@@ -11,6 +11,31 @@ use TenUpPlugin\Module;
 
 /**
  * Abstract class for post types.
+ *
+ *  Usage:
+ *
+ *  class FooPostType extends AbstractPostType {
+ *
+ *      public function get_name() {
+ *          return 'custom-post-type';
+ *      }
+ *
+ *      public function get_singular_label() {
+ *          return 'Custom Post'
+ *      }
+ *
+ *      public function get_plural_label() {
+ *          return 'Custom Posts';
+ *      }
+ *
+ *      public function get_menu_icon() {
+ *          return 'embed-post';
+ *      }
+ *
+ *      public function can_register() {
+ *          return true;
+ *      }
+ *  }
  */
 abstract class AbstractPostType extends Module {
 
--- a/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
+++ b/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
@@ -113,7 +113,7 @@ abstract class AbstractPostType extends Module {
 	 * @return array
 	 */
 	public function get_options() {
-		$options = [
+		return [
 			'labels'            => $this->get_labels(),
 			'public'            => true,
 			'has_archive'       => true,
@@ -126,8 +126,6 @@ abstract class AbstractPostType extends Module {
 			'menu_position'     => $this->get_menu_position(),
 			'hierarchical'      => $this->is_hierarchical(),
 		];
-
-		return $options;
 	}
 
 	/**
@@ -140,7 +138,7 @@ abstract class AbstractPostType extends Module {
 		$singular_label = $this->get_singular_label();
 
 		// phpcs:disable -- ignoring template strings without translators placeholder since this is dynamic
-		$labels = array(
+		$labels = [
 			'name'                     => $plural_label,
 			// Already translated via get_plural_label().
 			'singular_name'            => $singular_label,
@@ -169,7 +167,7 @@ abstract class AbstractPostType extends Module {
 			'item_updated'             => sprintf( __( '%s updated.', 'tenup-plugin' ), $singular_label ),
 			'menu_name'                => $plural_label,
 			'name_admin_bar'           => $singular_label,
-		);
+		];
 		// phpcs:enable
 
 		return $labels;
--- a/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
+++ b/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
@@ -36,6 +36,35 @@ abstract class AbstractPostType extends Module {
 	abstract public function get_plural_label();
 
 	/**
+	 * Get the menu icon for the post type.
+	 *
+	 * This can be a base64 encoded SVG, a dashicons class or 'none' to leave it empty so it can be filled with CSS.
+	 *
+	 * @see https://developer.wordpress.org/resource/dashicons/
+	 *
+	 * @return string
+	 */
+	abstract public function get_menu_icon();
+
+	/**
+	 * Get the menu position for the post type.
+	 *
+	 * @return int|null
+	 */
+	public function get_menu_position() {
+		return null;
+	}
+
+	/**
+	 * Is the post type hierarchical?
+	 *
+	 * @return bool
+	 */
+	public function is_hierarchical() {
+		return false;
+	}
+
+	/**
 	 * Default post type supported feature names.
 	 *
 	 * @return array
@@ -68,6 +97,9 @@ abstract class AbstractPostType extends Module {
 			'show_in_nav_menus' => false,
 			'show_in_rest'      => true,
 			'supports'          => $this->get_editor_supports(),
+			'menu_icon'         => $this->get_menu_icon(),
+			'menu_position'     => $this->get_menu_position(),
+			'hierarchical'      => $this->is_hierarchical(),
 		];
 
 		return $options;
--- a/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
+++ b/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
@@ -215,5 +215,4 @@ abstract class AbstractPostType extends Module {
 	public function after_register() {
 		// Do nothing.
 	}
-
 }
--- a/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
+++ b/mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
@@ -1,0 +1,187 @@
+<?php
+/**
+ * AbstractPostType
+ *
+ * @package TenUpPlugin
+ */
+
+namespace TenUpPlugin\PostTypes;
+
+use TenUpPlugin\Module;
+
+/**
+ * Abstract class for post types.
+ */
+abstract class AbstractPostType extends Module {
+
+	/**
+	 * Get the post type name.
+	 *
+	 * @return string
+	 */
+	abstract public function get_name();
+
+	/**
+	 * Get the singular post type label.
+	 *
+	 * @return string
+	 */
+	abstract public function get_singular_label();
+
+	/**
+	 * Get the plural post type label.
+	 *
+	 * @return string
+	 */
+	abstract public function get_plural_label();
+
+	/**
+	 * Default post type supported feature names.
+	 *
+	 * @return array
+	 */
+	public function get_editor_supports() {
+		$supports = [
+			'title',
+			'editor',
+			'author',
+			'thumbnail',
+			'excerpt',
+			'revisions',
+		];
+
+		return $supports;
+	}
+
+	/**
+	 * Get the options for the post type.
+	 *
+	 * @return array
+	 */
+	public function get_options() {
+		$options = [
+			'labels'            => $this->get_labels(),
+			'public'            => true,
+			'has_archive'       => true,
+			'show_ui'           => true,
+			'show_in_menu'      => true,
+			'show_in_nav_menus' => false,
+			'show_in_rest'      => true,
+			'supports'          => $this->get_editor_supports(),
+		];
+
+		return $options;
+	}
+
+	/**
+	 * Get the labels for the post type.
+	 *
+	 * @return array
+	 */
+	public function get_labels() {
+		$plural_label   = $this->get_plural_label();
+		$singular_label = $this->get_singular_label();
+
+		// phpcs:disable -- ignoring template strings without translators placeholder since this is dynamic
+		$labels = array(
+			'name'                     => $plural_label,
+			// Already translated via get_plural_label().
+			'singular_name'            => $singular_label,
+			// Already translated via get_singular_label().
+			'add_new_item'             => sprintf( __( 'Add New %s', 'tenup-plugin' ), $singular_label ),
+			'edit_item'                => sprintf( __( 'Edit %s', 'tenup-plugin' ), $singular_label ),
+			'new_item'                 => sprintf( __( 'New %s', 'tenup-plugin' ), $singular_label ),
+			'view_item'                => sprintf( __( 'View %s', 'tenup-plugin' ), $singular_label ),
+			'view_items'               => sprintf( __( 'View %s', 'tenup-plugin' ), $plural_label ),
+			'search_items'             => sprintf( __( 'Search %s', 'tenup-plugin' ), $plural_label ),
+			'not_found'                => sprintf( __( 'No %s found.', 'tenup-plugin' ), strtolower( $plural_label ) ),
+			'not_found_in_trash'       => sprintf( __( 'No %s found in Trash.', 'tenup-plugin' ), strtolower( $plural_label ) ),
+			'parent_item_colon'        => sprintf( __( 'Parent %s:', 'tenup-plugin' ), $plural_label ),
+			'all_items'                => sprintf( __( 'All %s', 'tenup-plugin' ), $plural_label ),
+			'archives'                 => sprintf( __( '%s Archives', 'tenup-plugin' ), $singular_label ),
+			'attributes'               => sprintf( __( '%s Attributes', 'tenup-plugin' ), $singular_label ),
+			'insert_into_item'         => sprintf( __( 'Insert into %s', 'tenup-plugin' ), strtolower( $singular_label ) ),
+			'uploaded_to_this_item'    => sprintf( __( 'Uploaded to this %s', 'tenup-plugin' ), strtolower( $singular_label ) ),
+			'filter_items_list'        => sprintf( __( 'Filter %s list', 'tenup-plugin' ), strtolower( $plural_label ) ),
+			'items_list_navigation'    => sprintf( __( '%s list navigation', 'tenup-plugin' ), $plural_label ),
+			'items_list'               => sprintf( __( '%s list', 'tenup-plugin' ), $plural_label ),
+			'item_published'           => sprintf( __( '%s published.', 'tenup-plugin' ), $singular_label ),
+			'item_published_privately' => sprintf( __( '%s published privately.', 'tenup-plugin' ), $singular_label ),
+			'item_reverted_to_draft'   => sprintf( __( '%s reverted to draft.', 'tenup-plugin' ), $singular_label ),
+			'item_scheduled'           => sprintf( __( '%s scheduled.', 'tenup-plugin' ), $singular_label ),
+			'item_updated'             => sprintf( __( '%s updated.', 'tenup-plugin' ), $singular_label ),
+			'menu_name'                => $plural_label,
+			'name_admin_bar'           => $singular_label,
+		);
+		// phpcs:enable
+
+		return $labels;
+	}
+
+	/**
+	 * Registers a post type and associates its taxonomies.
+	 *
+	 * @uses $this->get_name() to get the post's type name.
+	 * @return Bool Whether this theme has supports for this post type.
+	 */
+	public function register() {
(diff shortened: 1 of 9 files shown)


mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
```diff
@@ -0,0 +1,187 @@
...
+
+	/**
+	 * Registers a post type and associates its taxonomies.
+	 *
+	 * @uses $this->get_name() to get the post's type name.
+	 * @return Bool Whether this theme has supports for this post type.
+	 */
+	public function register() {
+		$this->register_post_type();
+		$this->register_taxonomies();
```
fabiankaegy, line 184: Should there also be a method to register additional meta fields for the CPT?

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
```diff
@@ -0,0 +1,187 @@
...
+
+namespace TenUpPlugin\PostTypes;
+
+use TenUpPlugin\Module;
+
+/**
+ * Abstract class for post types.
+ */
+abstract class AbstractPostType extends Module {
+
```
fabiankaegy, line 41: I think we need another method for getting the menu icon used by the CPT

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
```diff
@@ -0,0 +1,187 @@
...
+	/**
```
fabiankaegy, line 42: I think we need another method for getting the menu position used by the CPT

mu-plugins/10up-plugin/includes/classes/Taxonomies/AbstractTaxonomy.php
```diff
@@ -0,0 +1,142 @@
...
+
+	/**
+	 * Get the options for the taxonomy.
+	 *
+	 * @return array
+	 */
+	public function get_options() {
+		return array(
+			'labels'            => $this->get_labels(),
+			'hierarchical'      => false,
```
fabiankaegy: Should we make it easier to allow someone to switch between hierarchical and non-hierarchical?

fabiankaegy (commented):
Thanks for adding this :) I left some thoughts inline. 

Also just so you are aware, I've been working on a similar PR in the UI KIt Scaffold. https://github.com/10up/ui-kit-scaffold/pull/8

Whatever we come up with should be shared between the two

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
darylldoyle: Good call!

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
darylldoyle: @fabiankaegy so the abstract post type already has an `after_register()` method that can be used to do anything you need after the post type has been registered, including registering meta fields. Do you think that, with some documentation would suffice, or are you thinking a dedicated meta registration method?

If the latter, how are you picturing it working? 

The other option is that we could utilise a PHP trait for meta registration that would allow us to share the meta registration method across post types and taxonomies, whilst sharing a common API 🤔

fabiankaegy: I'm not sure what the best answer is here. I recognize that meta isn't necessarily easy because as you mention it may need to be registered for multiple post types etc. 

I'd almost like to come up with a way to handle meta registration in a separate way kind of like taxonomies or even blocks. Because in the end a piece of meta really is both registering the field but then also creating the UI for it likely in React these days. So it's kind or it's own module (part of a feature)... 

I'm not even sure we need to standardize it now. But I see many projects doing it in many different ways which I think eliminating this is one of the goals we have here. 

We can merge this with some docs for how you do it now and then maybe revisit in the future :) 

My goal is it for us to make it so easy that we finally can convince folks to stop loading field manager =D

darylldoyle: @fabiankaegy I _really_ like the idea of meta being shipped with the JS needed to register it. I wonder if there's some way that we can build some default react interfaces that can be extended 🤔 

Can you open an issue that we can hash these ideas out in and we'll leave this PR as-is for now. I don't want to over-complicate it too much or make it _too_ abstract before we've got anything merged.

I'll add docs around the usage of this code and how to add post meta etc 🙂

darylldoyle:
@fabiankaegy I've updated this PR and added a first-draft of the docs for adding post types. Can you please review?

fabiankaegy (approved):
This is great :) 

Let's ship this as is. Long term I would love to remove the `Demo.php` file from the scaffold and instead have some code generators that allow you to add them.

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
```diff
@@ -0,0 +1,218 @@
...
+	 * Get the post type name.
+	 *
+	 * @return string
+	 */
+	abstract public function get_name();
```
claytoncollie, line 47: @darylldoyle Seeing as how the currently supported PHP version is 8.2, should we use parameter and return type hints for all methods instead of only in the doc blocks? I am assuming that people using this scaffold is on new projects and most likely running on the latest version of PHP at the start of their project.

mu-plugins/10up-plugin/includes/classes/PostTypes/Post.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @package TenUpPlugin
+ */
+
+namespace TenUpPlugin\PostTypes;
+
+/**
+ * Post Post type.
+ */
+class Post extends \TenUpPlugin\Module {
```
claytoncollie: @darylldoyle On the AbstractTaxonomy class, you are using the `use` keyword for importing the Module class but are not in this file or in the Page.php file. Should they all use a single pattern?

https://github.com/10up/wp-scaffold/pull/218/files#diff-6c212e20a3d927335e7dac31df758af53b2e12e7134fc3c8273177c0df1d9040R10

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * @package TenUpPlugin
+ */
+
+namespace TenUpPlugin\PostTypes;
+
+use TenUpPlugin\Module;
+
+/**
+ * Abstract class for post types.
```
claytoncollie, line 13: @darylldoyle This doc blck is missing the implementation documentation that is available in the AbstractTaxonomy class doc block. 

https://github.com/10up/wp-scaffold/pull/218/files#diff-6c212e20a3d927335e7dac31df758af53b2e12e7134fc3c8273177c0df1d9040R15

mu-plugins/10up-plugin/includes/classes/Taxonomies/AbstractTaxonomy.php
```diff
@@ -0,0 +1,162 @@
...
+		return true;
+	}
+
+	/**
+	 * Get the options for the taxonomy.
+	 *
+	 * @return array
+	 */
+	public function get_options() {
+		return array(
```
claytoncollie: @darylldoyle This method is following a different pattern compared the AbstractPostType where `$options` is defined at the top of the method and then returned at the bottom. I personally like this pattern better but whatever it is it should be consistent.

mu-plugins/10up-plugin/includes/classes/Taxonomies/Demo.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
claytoncollie, line 1: @darylldoyle Similar to what Fabian said, these Demo.php files might not be neeed when you have the markdown file explaining how to use the classes as well as the doc block example within each abstract class file.

https://github.com/10up/wp-scaffold/pull/218/files#diff-6c212e20a3d927335e7dac31df758af53b2e12e7134fc3c8273177c0df1d9040R15

mu-plugins/10up-plugin/includes/classes/PostTypes/Post.php
darylldoyle: Good call!

mu-plugins/10up-plugin/includes/classes/PostTypes/AbstractPostType.php
darylldoyle: While I agree with this myself, the rest of the codebase isn't using return type hints, so it feels odd to force them here. I think it'd be better to merge without return type hints and then do a wholesale update to the scaffold to add them to :all-the-things:

mu-plugins/10up-plugin/includes/classes/Taxonomies/Demo.php
darylldoyle: Agreed, the other side of it is that they're not registered by default, so won't affect sites. They also provide a very easy copy/paste solution until we can get a CLI scaffold script created. At that point, I think it'd be worth removing them

dsawardekar:
Great job on this, @darylldoyle! Thanks for putting it together.

A point that might need further clarification is the inclusion of the Post & Page classes. This has been a frequent area of confusion, given the existing Post and Page post types in core. The addition here might seem redundant to some, so expanding on their purpose and functionality could help clear up any misunderstandings.

The class hierarchy for these classes seems incorrect. Maybe we should extend the AbstractPostType, and add no-op with comment(s). That way it is clearer that these are not stub classes, and customization to Core Page / Post post types should go here.

nicholasio:
@darylldoyle Could we add a more structured way for registering custom fields? I'd love to use register_rest_field and/register_meta registering custom fields in this way where you also declare the "schema" makes things easier for headless projects. I'm thinking mostly from a REST API standpoint and not really thinking about us creating a wrapper around rendering the custom fields in the editor.

The end goal here is to make sure we're consistent in how we are exposing custom fields through the REST API. There are lots of ways to do it and it doesn't always get added to the REST API schema.

See this for example: https://github.com/10up/headstartwp/blob/develop/wp/headless-wp/includes/classes/Preview/PreviewLink.php#L36

darylldoyle:
@dsawardekar I'd love your eyes on the latest update I've pushed. It introduces an `AbstractCorePostType` class that extends `AbstractPostType` and kills any methods not needed by the core post types. That allows us to maintain the same interface for taxonomy registration and the `after_register()` method.

@nicholasio I've spoken about this topic at length with @ivanlopez on our 1:1's. There are a bunch of ways to do it, we just need to find a way to handle it in a way that's both flexible and as simple as possible for smaller projects. Can we continue this discussion in https://github.com/10up/wp-framework/issues/6, please? I want to keep this PR solely to the registration of post types and taxonomies 🙂

dsawardekar:
@darylldoyle I like the `after_register`, it will make things a lot more predictable across projects. Looks good to me.

This might be on your radar. This PR opens the door for integrating scaffolding commands that can rapidly bootstrap new Post Types and Taxonomies. :)

theskinnyghost:
I'm here after reading the internal blog post and I was wondering if you were aware of [the "Extended CPTs" library](https://github.com/johnbillion/extended-cpts) created by the core committer John Blackbourn. 
Basically it extends the default WP functions to register CPTs and Taxonomies to include additional parameters to make things like adding custom admin columns, filters, etc.. much easier. I've used it in a few projects in the past and it works great!
I thought I would share it here in case you'd be interested in using it in these abstract classes.

darylldoyle:
Thanks for flagging that @theskinnyghost; I'd not seen that before.

Whilst I think it's a great concept, I think it does _too much_ to be a core part of the scaffold. However, I would love to see us provide ways to pull in parts of what it does. It'll definitely be a great source of inspiration.

@dsawardekar that scaffold command is definitely on my radar and is something that @fabiankaegy has already been working on for blocks. I'd love to see us have something akin to `artisan` in Laravel: 

```shell
wp 10up create post-type report --hierarchical
```

PR #223, opened 2024-03-19
Fix/phpcs

### LINKED ISSUE #198: PHPCS does not scan mu-plugins folder

### LINKED ISSUE #219: PHPCS Broken
### Describe the bug

Since #206 was merged, PHPCS has been broken on commit. This is because it moved the lintstaged config to the theme & plugin and they don't contain the PHPCS config, nor the executable. 

We should either:

1. Find a way to allow PHP linting to run from the root, whilst allowing eslint to use the configs from each directory
2. Revert the lintstaged change for now
3. Install and configure PHPCS and it's rules within each directory (but this adds a lot of config overhead)

### Steps to Reproduce

* Clone the repo.
* Make a change within a PHP file.
* Commit.
* See PHPCS error in the pre-commit hook.

### Code of Conduct

darylldoyle:
### Description of the Change
This PR does a couple of things:

* Fixes an issue where PHPCS was not running when changes were made in the theme or plugin
* Upgrades the 10up PHPCS to the latest version
* Fixes current PHPCS issues
* Adds a PHPCS check to Github Actions within the scaffold

Closes #219 
Closes #216 
Closes #198 

### How to test the Change
* Run PHPCS

### Credits

Props @darylldoyle @Jamesking56

### Checklist:

### DIFF
--- a/mu-plugins/10up-plugin/plugin.php
+++ b/mu-plugins/10up-plugin/plugin.php
@@ -1,4 +1,20 @@
 <?php
+/**
+ * Plugin Name:       10up Plugin Scaffold
+ * Description:       A brief description of the plugin.
+ * Version:           0.1.0
+ * Requires at least: 4.9
+ * Requires PHP:      7.2
+ * Author:            10up
+ * Author URI:        https://10up.com
+ * License:           GPL v2 or later
+ * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+ * Text Domain:       tenup-plugin
+ * Domain Path:       /languages
+ * Update URI:        https://github.com/10up/wp-scaffold
+ *
+ * @package           TenUpPlugin
+ */
 
 // Useful global constants.
 define( 'TENUP_PLUGIN_VERSION', '0.1.0' );
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -12,6 +12,11 @@
 
 	<rule ref="10up-Default" />
 
+	<!-- Ignore filecomment for the plugin loader -->
+	<rule ref="Squiz.Commenting.FileComment.Missing">
+		<exclude-pattern>./mu-plugins/10up-plugin-loader.php</exclude-pattern>
+	</rule>
+
 	<arg value="sp"/> <!-- Show sniff and progress -->
 	<arg name="colors"/> <!-- Show results with colors. Disable if working on Windows -->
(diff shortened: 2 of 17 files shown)


mu-plugins/10up-plugin/plugin.php
```diff
@@ -1,4 +1,20 @@
...
+ * Requires at least: 4.9
+ * Requires PHP:      7.2
+ * Author:            10up
+ * Author URI:        https://10up.com
+ * License:           GPL v2 or later
+ * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+ * Text Domain:       tenup-plugin
+ * Domain Path:       /languages
+ * Update URI:        https://github.com/10up/wp-scaffold
+ *
```
fabiankaegy, line 15: Isn't all this metadata that never actually shows anywhere since mu-plugins are using the plugin loader approach?

phpcs.xml
```diff
@@ -1,9 +1,24 @@
...
 	<description>10up PHPCS extended.</description>
+
+	<!-- Scan these directories -->
+	<file>./themes/*</file>
+	<file>./mu-plugins/*</file>
+
+	<!-- Don't scan these directories -->
 	<exclude-pattern>dist/</exclude-pattern>
-	<exclude-pattern>plugins/*</exclude-pattern>
+	<exclude-pattern>./plugins/*</exclude-pattern>
```
fabiankaegy: I feel like this will cause issues pretty soon. To me there are valid reasons why I may need to add custom plugins to the plugins directory and 3rd party plugins to the mu-plugins folder. 

Maybe it makes more sense to not use any wildcards but explicitly specify the folders we want?

fabiankaegy (approved):
Left two suggestions / questions. But this is great :)

ivanlopez (approved):

mu-plugins/10up-plugin/plugin.php
darylldoyle: Correct, though it's not uncommon for someone to copy-paste the mu-plugin as a scaffold for a standard plugin, in which case it's helpful to have this in place.

phpcs.xml
darylldoyle: I'm happy with that. Updated to exclude `node_modules` and `vendor` but include `plugins` by default

PR #229, opened 2024-05-20
Add logic for block style entry points

dmtrmrv:
This PR introduces the `enqueue_block_specific_styles()` function in the `blocks.php` file. This function is designed to enqueue styles that are specific to a block.

The function operates by first retrieving all the CSS files in the 'blocks/autoenqueue' directory. For each stylesheet, it determines the block type by removing the directory path and '.css' from the stylesheet path.

The function then attempts to get the asset file for the block type. If the asset file doesn't exist, it creates a new one with the version set to the file modification time of the stylesheet and no dependencies.

Finally, the function enqueues the block style using the block type, the URL to the stylesheet, the path to the stylesheet, the version from the asset file, and the dependencies from the asset file.

This is required for https://github.com/10up/10up-toolkit/issues/339

## How To Test
Once https://github.com/10up/10up-toolkit/pull/390 is merged and released, the /dist/blocks/ directory will contain a new folder: /autoenqueue/. I was able to test this by generating block styles via the 10up-toolkit repository and then copying the /autoenqueue/ folder into the scaffold.

## Credits
Props @fabiankaegy, @dmtrmrv, @Antonio-Laguna

### Checklist:


fabiankaegy:
Marking this as a draft till the toolkit PR has been merged & released :)

themes/10up-theme/includes/blocks.php
```diff
@@ -128,6 +130,52 @@ function blocks_editor_styles() {
...
+		wp_enqueue_block_style(
+			$block_type,
+			[
+				'handle'       => "{$block_type}",
+				'src'          => TENUP_THEME_DIST_URL . 'blocks/autoenqueue/' . $block_type . '.css',
+				'path'         => $stylesheet_path,
+				'version'      => $asset_file['version'],
+				'dependencies' => $asset_file['dependencies'],
+			]
+		);
```
fabiankaegy: One issue I ran into whilst testing this on a project is that `wp_enqueue_block_style` only registers the blocks stylesheet when the block is actually used on a page. 

This seems correct at first. However this means you are not able to add the handle of this block style as a stylesheet dependency of another custom block for example. 

If you need to load the custom stylesheet of the `core/list` block for a custom block that also needs the list styles you would want to do the following:

```json
{
    "styles": [ "file:./style.css", "namespace-theme-core/list-styles" ]
}
```

The way to accomplish this is to split the enque action from the registration. 

```suggestion
		wp_register_style(
				"{$block_type}",
				TENUP_THEME_DIST_URL . 'blocks/autoenqueue/' . $block_type . '.css',
				$asset_file['version'],
				$asset_file['dependencies'],
		);

		wp_enqueue_block_style(
			$block_type,
			[
				'handle'       => "{$block_type}",
				'path'         => $stylesheet_path,

			]
		);
```

themes/10up-theme/includes/blocks.php
```diff
@@ -128,6 +130,52 @@ function blocks_editor_styles() {
...
+			$asset_file = [
+				'version'      => filemtime( $stylesheet_path ),
+				'dependencies' => [],
+			];
+		}
+
+		wp_enqueue_block_style(
+			$block_type,
+			[
+				'handle'       => "{$block_type}",
```
fabiankaegy: Another change I'd recommend is changing the handle here to something that is namespace with the theme.

If we could make it so the handle is something like: `namespace-theme-${block_namespace}-${block_name}-style`

fabiankaegy: Resolved in [290815e](https://github.com/10up/wp-scaffold/pull/229/commits/290815e423c7a71afe3f2047a97d3f4bf7045f49)

themes/10up-theme/includes/blocks.php
fabiankaegy: Resolved in [290815e](https://github.com/10up/wp-scaffold/pull/229/commits/290815e423c7a71afe3f2047a97d3f4bf7045f49)

fabiankaegy:
This PR has been updated and tested  👍 Once toolkit 6.3 gets officially released we can merge this in  👍

fabiankaegy (approved):

PR #236, opened 2024-08-13
Feature/updated autoloading

darylldoyle:
### Description of the Change
This PR updates the class autoloader from using https://gitlab.com/hpierce1102/ClassFinder/ to using https://github.com/spatie/php-structure-discoverer and adds unit tests to ensure it's working as expected.

Spatie are a well known company, who specialise in Open Source packages. This means the package is updated more often and bugs will be fixed more quickly. For example, the last commit in `hpierce1102/classfinder` was June 18, 2023, whereas the last commit in `spatie/php-structure-discoverer` was today.

There are a few other benefits of the Spatie package:

* It doesn't rely on the `composer.json` to load files, meaning we no longer have to deploy that.
* There's no hidden logic; we can see exactly how the classes are being loaded now and what constraints are being put on them.
* We can conditionally cache the class loader, giving it a considerable performance boost on production/staging.

**Old Class Loader**
Screenshot 2024-08-13 at 23 22 45

**New Class Loader**
Screenshot 2024-08-13 at 23 21 50

Additionally, during the refactor, I eliminated all the hardcoded strings and awkward configurations needed to get the classloader working. It should now be plug-and-play.

One downside of this change is that it only supports PHP 8.1 upwards, though since every version pre-8.1 is End of Life, I think that's a fair tradeoff, https://www.php.net/supported-versions.php.

### How to test the Change
Ensure all classes are still loading and test suite is passing.

### Changelog Entry
* Changed - Class loading swapped out for a newer, more performant library.

### Credits

Props @darylldoyle 

### Checklist:


bin/install-wp-tests.sh
darylldoyle, line 1: This is an updated version of the default `install-wp-tests.sh`, it has a few benefits:

- It has a help command, which ensures engineers that are new to PHPUnit have a way to understand the params of the script
- When run from a Local WP shell, it automatically picks up and uses the MySQL socket.
It has a sensible set of defaults that will work for 99% of 10up builds, meaning you can run the script without passing any parameters.
- It resolves an issue where the install script thinks it's already installed (because the `/tmp` directory exists, but the DB isn't installed.

themes/10up-theme/tests/includes/classes/ModuleInitializationTest.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @return void
+	 */
+	public function test_it_can_find_classes() {
+		$this->class->init_classes();
+
+		$class  = new \ReflectionClass( $this->class );
+		$method = $class->getMethod( 'get_classes' );
+
+		$classes = $method->invoke( $this->class );
+		$this->assertCount( 2, $classes );
```
tobeycodes: I think this should be a count greater than 1. If it can find 1 then it's working. Otherwise you need to update this test every time you include another class.

darylldoyle: Good call, I had this on a project that I'm using this on. I'll update with the tests I'm using there!

mu-plugins/10up-plugin/includes/classes/ModuleInitialization.php
```diff
@@ -56,9 +58,27 @@ private function __construct() {
...
+		$class_finder = Discover::in( __DIR__ );
+		// Only fetch classes.
+		$class_finder->classes();
+		// Only fetch classes in the current namespace.
+		$class_finder->custom(
+			fn( DiscoveredStructure $structure ) => str_starts_with( $structure->namespace, __NAMESPACE__ )
+		);
+
+		// If we are in production or staging, cache the class loader to improve performance.
+		if ( in_array( wp_get_environment_type(), [ 'production', 'staging' ], true ) ) {
```
tobeycodes, line 71: **thought:** i think we should assume production and always cache. Only disable when local is explicilty set. In theory WP hosts could use anything for this value and would not count all possible env types like preprod, uat, qa, etc that some proojects use

darylldoyle: @tobeycodes `wp_get_environment_type()` only has 4 available options, using anything other than those will default to production. See: https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/load.php#L229-L234

tobeycodes: Oh... I should have read the source code before commenting. This makes sense then

darylldoyle:
Closing this out since it's now handled by the framework

PR #238, opened 2024-09-30
Add a block based theme option

fabiankaegy:
### Description of the Change

Adds a new Block Theme scaffold to the repository

### Changelog Entry

Added - A new Block-Based Theme Scaffold

### Credits

Props @fabiankaegy @tobeycodes @darylldoyle @colinswinney


fabiankaegy:
CC: @10up/editorial-engineering Would love to get your all's input on this :) 

Also CC: @darylldoyle we can use this new scaffold to update how we are doing things. So if we want to introduce larger coding style / structural changes this is a time where we can include stuff like that :)

themes/10up-block-theme/webpack.config.js
fabiankaegy, line 1: This is no longer needed now that Toolkit 6.3 is out. We can safely remove this :)

themes/10up-block-theme/theme.json
```diff
@@ -0,0 +1,390 @@
+{
+	"settings": {
+		"appearanceTools": true,
+		"color": {
+			"defaultDuotone": false,
+			"defaultGradients": false,
+			"defaultPalette": false,
+			"palette": [
```
fabiankaegy: Pretty much all the tokens in here are inspired by UI Kit. I don't think we should be as opinionated in the scaffold. 

Though I would like us to ship stuff like the section styles for color theming.

themes/10up-block-theme/stylelint.config.js
fabiankaegy, line 1: We should double check why this file needed to be created and patch it upstream in toolkit

themes/10up-block-theme/screenshot.png
fabiankaegy, line 1: Needs to be removed

themes/10up-block-theme/phpcs.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<ruleset name="10up PHPCS">
+	<description>10up PHPCS extended.</description>
+	<exclude-pattern>dist/</exclude-pattern>
+	<exclude-pattern>plugins/*</exclude-pattern>
```
fabiankaegy: Why is this here? 

Also do we even need this file if there is one at the root of the project?

themes/10up-block-theme/includes/utility.php
```diff
@@ -0,0 +1,79 @@
...
+ *    Optional. Array of arguments for enqueuing a view script.
+ *    @type string $handle Script handle.
+ *    @type string $src Script URL.
+ *    @type array $dep Script dependencies.
+ *    @type string|bool $ver Script version.
+ *    @type bool $in_footer Whether to enqueue the script before </body> instead of in the <head>.
+ * }
+ * @return void
+ */
+function wp_enqueue_block_view_script( $block_name, $args = array() ) {
```
fabiankaegy: If we want to include this Utility function we should change the name spacing to ensure this doesn't break if WP actually ships this utility.

themes/10up-block-theme/includes/overrides.php
```diff
@@ -0,0 +1,84 @@
...
+ * @link https://developer.wordpress.org/reference/functions/print_emoji_styles/
+ * @link https://developer.wordpress.org/reference/functions/wp_staticize_emoji/
+ * @link https://developer.wordpress.org/reference/functions/wp_staticize_emoji_for_email/
+ * @link https://developer.wordpress.org/reference/functions/wp_generator/
+ * @link https://developer.wordpress.org/reference/functions/wlwmanifest_link/
+ * @link https://developer.wordpress.org/reference/functions/rsd_link/
+ *
+ * @return void
+ */
+function setup() {
```
fabiankaegy, line 23: Is all of this still relevant today? Let's audit!

themes/10up-block-theme/includes/core.php
```diff
@@ -0,0 +1,190 @@
...
...t:"The async attribute is redundant on inline scripts"}head script:not([src])[defer]::before{content:"The defer attribute is redundant on inline scripts"}head [rel=stylesheet][href^="//"],head [rel=stylesheet][href^=http],head script[src][src][src^="//"],head script[src][src][src^=http]{border-style:var(--ct-is-problematic);border-color:var(--ct-error)}head script[src][src][src^="//"]::before,head script[src][src][src^=http]::before{content:"[Third Party Blocking Script – " attr(src) "]"}head [rel=stylesheet][href^="//"]::before,head [rel=stylesheet][href^=http]::before{content:"[Third Party Blocking Stylesheet – " attr(href) "]"}head script~meta[http-equiv=content-security-policy]{border-style:var(--ct-is-problematic);border-color:var(--ct-error)}head script~meta[http-equiv=content-security-policy]::before{content:"[Meta CSP defined after JS]"}head>meta[charset]:not(:nth-child(-n+5)){border-style:var(--ct-is-problematic);border-color:var(--ct-warn)}head>meta[charset]:not(:nth-child(-n+5))::before{content:"[Charset should appear as early as possible]"}link[rel=stylesheet].ct,link[rel=stylesheet][media=print],script[async],script[defer],script[type=module],style.ct{display:none}' );
```
fabiankaegy: @dainemawer are we actually using this?

themes/10up-block-theme/includes/core.php
```diff
@@ -0,0 +1,190 @@
...
+ * Handles scrollbar width detection.
+ *
+ * Adds a JavaScript event listener to the DOMContentLoaded event. When the DOM is fully loaded,
+ * it calculates the width of the scrollbar and sets a CSS variable `--wp--custom--scrollbar-width` with the width.
+ * It also adds an event listener to the window resize event to update the scrollbar width when the window is resized.
+ *
+ * @return void
+ */
+function scrollbar_detection() {
+	echo '<script>window.addEventListener("DOMContentLoaded",()=>{const t=()=>window.innerWidth-document.body.clientWidth;const e=()=>{document.documentElement.style.setProperty("--wp--custom--scrollbar-width",`${t()}px`)};e();window.addEventListener("resize",e);});</script>' . "\n";
```
fabiankaegy: Not sure we want this in the scaffold.

themes/10up-block-theme/includes/core.php
```diff
@@ -0,0 +1,190 @@
...
+	);
+
+	\UIKitCore\Helpers\register_icons(
+		[
+			'name'  => 'tenup',
+			'label' => 'Theme Icons',
+			'icons' => $icons,
+		]
+	);
+}
```
fabiankaegy, line 134: Very torn if we want this here. 

I think every single project we build needs to have a solution for icons. But this makes it dependent on UI Kit. So not sure what the best option is.

joesnellpdx: @fabiankaegy 

I'd keep it. I feel like this makes sense - if we are using UI Kit - "it should just work" and your early escape resolves if not. 

I would add more detail in comments regarding the dependency on UI Kit. 

If we find helpful, someone can add an issue to utilize a more agnostic approach option within this function.

themes/10up-block-theme/includes/core.php
joesnellpdx: [Here is more context](https://fueled10up.slack.com/archives/C10D34T33/p1680624401811279) 

CC @dmtrmrv

themes/10up-block-theme/stylelint.config.js
joesnellpdx: Might give some context: https://fueled10up.slack.com/archives/C035LFXLQH0/p1692947406604129

@fabiankaegy

joesnellpdx:
@fabiankaegy 

I'll also share with the FEE Best Practices WG - as some of the suggested and similar changes here have been discussed in that WG

themes/10up-block-theme/assets/css/base/link.css
```diff
@@ -0,0 +1,11 @@
+/*
+ * Ensure that we are not overriding the editor styles of the
+ * WordPress Components.
+ *
+ * This is a workaround for: https://github.com/WordPress/gutenberg/issues/10178
+ *
+ * using :where to prevent the specificity increase of using :not
+ */
+a:where(:not(.components-external-link, :has(> img:only-child), .wp-element-button)) {
```
colinswinney: I would like to see a similar file for form elements.  I've seen a few projects that don't scope their input/select etc selectors and it affects the editor placeholder components. I've used the following with success to prevent bleeding into those. 

```css
// base/forms.css
&:where(:not(.components-placeholder)) {

    input {...}
}
```

themes/10up-block-theme/assets/css/base/reset.css
```diff
@@ -0,0 +1,85 @@
+/*
```
colinswinney, line 1: I've been adding this to my resets as well to ensure alt tags remain readable if the image doesn't load.  Sometimes you see folks put `font-size: 0;` or `line-height: 0` on figure/div tags to avoid that small bit of white space that can occur, this protects against that situation. 

```css
img {
  font-size: 1rem;
  font-style: italic;
  line-height: 1.5;
}
```

themes/10up-block-theme/assets/css/mixins/margin-collapse.css
```diff
@@ -0,0 +1,14 @@
+@define-mixin margin-collapse {
```
colinswinney: I would recommend calling this `margin-trim` to better reflect what its doing.

themes/10up-block-theme/assets/css/utilities/visually-hidden.css
```diff
@@ -0,0 +1,10 @@
+.visually-hidden {
```
colinswinney, line 1: I would add this to the mixins and then call that here.  It's very handy to have available in your styles vs only as a utility class.

```css
@define-mixin visually-hidden {
        border: 0;
	clip: rect(0, 0, 0, 0);
	height: 1px;
	margin: -1px;
	overflow: hidden;
	padding: 0;
	position: absolute;
	width: 1px;
}

.visually-hidden {
        @mixin visually-hidden;
}
```

themes/10up-block-theme/assets/css/base/layout.css
```diff
@@ -0,0 +1,73 @@
...
+:root {
+	--full-viewport-width: var(--wp--custom--full-viewport-width);
+	--site-outer-padding: var(--wp--custom--site-outer-padding);
+	--site-content-width: var(--wp--custom--site-content-width);
+	--main-content-width-side-spacing: var(--wp--custom--main-content-width-side-spacing);
+	--main-wide-width-side-spacing: var(--wp--custom--main-wide-width-side-spacing);
+	/* stylelint-disable-next-line length-zero-no-unit */
+	--site-header-top-offset: var(--wp-admin--admin-bar--height, 0px);
+
+	&.has-pinned-header {
```
colinswinney: Is `has-pinned-header` a class generated by WP in any way?  This seems a bit unnecessary unless there is some mechanism by which that would be commonly added in addition to the `--header-height` var.

fabiankaegy: Whoops, yeah these should be removed. They are project specific

themes/10up-block-theme/assets/css/utilities/visually-hidden.css
fabiankaegy: I do like that pattern but I also worry that it leads to a lot of duplicated / bloated CSS 🤔

themes/10up-block-theme/assets/css/mixins/margin-collapse.css
fabiankaegy: Agreed! :)

themes/10up-block-theme/includes/core.php
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Set up theme defaults and register supported WordPress features.
+ *
+ * @return void
+ */
+function setup() {
+	$n = function ( $function_name ) {
+		return __NAMESPACE__ . "\\$function_name";
+	};
```
darylldoyle: **thought (non-blocking):** It seems like this is unused now (🤞) Let's remove it anyway and use the full namespace as we are below in all-the-places

themes/10up-block-theme/functions.php
```diff
@@ -0,0 +1,40 @@
...
+require_once TENUP_BLOCK_THEME_INC . 'core.php';
+require_once TENUP_BLOCK_THEME_INC . 'blocks.php';
+require_once TENUP_BLOCK_THEME_INC . 'overrides.php';
+require_once TENUP_BLOCK_THEME_INC . 'template-tags.php';
+require_once TENUP_BLOCK_THEME_INC . 'utility.php';
+
+// Run the setup functions.
+TenupBlockTheme\Core\setup();
+TenupBlockTheme\Blocks\setup();
+TenupBlockTheme\TemplateTags\setup();
```
darylldoyle, line 40: **thought (non-blocking):** Conceptually, I have an issue with this approach, and it's not something we only do here. I don't understand why we mix OOP and functional code for no real reason. IMO, we should stick to one approach or another unless functions are being provided for templating reasons (in which case, we wouldn't need to `setup()` or `init()` any hooks; they should be encapsulated in classes.

themes/10up-block-theme/phpcs.xml
darylldoyle: This looks incorrect, based on the fact that it's excluding `plugins/*`, _but_, I do think there's some benefit to this, especially if we move ahead with something like https://github.com/10up/10up-toolkit/issues/424, as it would mean any themes/plugins pulled in via that command would have valid PHPCS configs, even if pulled into legacy codebases.

themes/10up-block-theme/includes/core.php
fabiankaegy: Yeah this should be removed. I used the full namespace so we get better IDE integration (Props @tobeycodes )

themes/10up-block-theme/functions.php
fabiankaegy: I'm all ear on what we can do to avoid this :)

themes/10up-block-theme/.lintstagedrc.json
```diff
@@ -0,0 +1,14 @@
+{
+    "*.css": [
+      "10up-toolkit lint-style"
+    ],
+    "*.js": [
+      "10up-toolkit lint-js"
+    ],
+    "*.jsx": [
+      "10up-toolkit lint-js"
+    ],
```
tobeycodes: **question:** should we add ts and tsx here?

themes/10up-block-theme/.lintstagedrc.json
```diff
@@ -0,0 +1,14 @@
...
+    "*.php": [
+      "./vendor/bin/phpcs --extensions=php --warning-severity=8 -s"
```
tobeycodes: **suggestion:** these arguments should in the config itself to prevent inconsistencies with IDE integration and lintstagedrc

```
    <fileextensions>
        <extension>php</extension>
    </fileextensions>

    <!-- Set the severity level for warnings -->
    <arg name="warning-severity" value="8"/>

    <!-- Show sniff codes in output (-s option) -->
    <arg name="s"/>
```

themes/10up-block-theme/.eslintrc
tobeycodes, line 1: **thought:** we should add this to the .editorconfig file and make sure it's formatted like the other rc files.

https://github.com/10up/wp-scaffold/blob/trunk/.editorconfig#L10

themes/10up-block-theme/.nvmrc
tobeycodes, line 1: **question:** can we bump this to the latest LTS?

themes/10up-block-theme/assets/css/base/layout.css
tobeycodes, line 1: **question:** why do we need the backwards compatibility? if a new project is started they won't be referencing the old variables and existing projects are not backwards porting future wp-scaffold updates

themes/10up-block-theme/composer.json
```diff
@@ -0,0 +1,31 @@
...
+  "name": "10up/tenup-theme",
+  "type": "wordpress-theme",
+  "authors": [
+    {
+      "name": "10up",
+      "email": "info@10up.com"
+    }
+  ],
+  "require": {
+    "php": ">=8.0",
```
tobeycodes: **question:** can we bump this?

themes/10up-block-theme/functions.php
tobeycodes: I agree with this but perhaps we should apply this to the whole scaffold in a separate PR. It'll be a big change

themes/10up-block-theme/includes/blocks.php
tobeycodes, line 1: **question:** should we be using a cache for performance (e.g. https://github.com/roots/bedrock-autoloader/blob/master/src/Autoloader.php) and perhaps RecursiveDirectoryIterator instead of glob?

themes/10up-block-theme/includes/blocks.php
```diff
@@ -0,0 +1,91 @@
...
+function register_theme_blocks() {
+	// Register all the blocks in the theme.
+	if ( file_exists( TENUP_BLOCK_THEME_BLOCK_DIST_DIR ) ) {
+		$block_json_files = glob( TENUP_BLOCK_THEME_BLOCK_DIST_DIR . '*/block.json' );
+
+		foreach ( $block_json_files as $filename ) {
+			$block_folder = dirname( $filename );
+			$block        = register_block_type_from_metadata( $block_folder );
+
+			add_filter(
```
tobeycodes: **request:** can this go outside of the foreach loop? here we are calling add_filter for every block registered rather than once with all the blocks to add

themes/10up-block-theme/includes/blocks.php
```diff
@@ -0,0 +1,91 @@
...
+ */
+function enqueue_theme_block_styles() {
+	$stylesheets = glob( TENUP_BLOCK_THEME_DIST_PATH . '/blocks/autoenqueue/**/*.css' );
+	foreach ( $stylesheets as $stylesheet_path ) {
+		$block_type = str_replace( TENUP_BLOCK_THEME_DIST_PATH . '/blocks/autoenqueue/', '', $stylesheet_path );
+		$block_type = str_replace( '.css', '', $block_type );
+		$asset_file = TENUP_BLOCK_THEME_DIST_PATH . 'blocks/autoenqueue/' . $block_type . '.asset.php';
+
+		if ( ! file_exists( $asset_file ) ) {
+			$asset_file = require $asset_file;
```
tobeycodes: **thought:** it appears there is a logical issue here. if the file does not exist, you can't require it

themes/10up-block-theme/includes/core.php
tobeycodes: Thank you! <3 IDE integration

tobeycodes (commented):
@fabiankaegy I saw this PR in my notifications and thought I'd leave some comments. Feel free to disregard anything that is not relevant. It looks great! Excited for block themes

themes/10up-block-theme/includes/core.php
fabiankaegy: Chatted with @dainemawer and agreed to remove it

themes/10up-block-theme/assets/css/base/link.css
fabiankaegy: @colinswinney I wonder if this issue has gotten less bad since we now have the iframed editor in most places? 

Though the more I think about it the more it is clear that any `Placeholders` still get rendered inside the iframe 🤔

themes/10up-block-theme/.lintstagedrc.json
fabiankaegy: Fixed

themes/10up-block-theme/assets/css/base/layout.css
fabiankaegy: Good point 👍

themes/10up-block-theme/composer.json
fabiankaegy: Done

themes/10up-block-theme/includes/blocks.php
fabiankaegy: Done

themes/10up-block-theme/includes/blocks.php
fabiankaegy: Fixed 👍

fabiankaegy:
This should be ready for another set of reviews :) 

Would love to get it merged soon in the new year 🚀

tobeycodes (approved):
This looks good to me. I think anything minor we can follow up in further PRs

themes/10up-block-theme/includes/blocks.php
darylldoyle: I agree here. @fabiankaegy can we get an issue to follow up on this? It's not block theme dependent, so no need to block this PR

themes/10up-block-theme/style.css
```diff
@@ -0,0 +1,10 @@
+/**
+ * Theme Name:  TENUP Block Theme
+ * Theme URI:   https://10up.com
+ * Description: A theme for the TENUP project.
```
darylldoyle: **nitpick (non-blocking):** Should this be `TENUP` or `TenUp`?

darylldoyle (approved):

PR #250, opened 2024-11-12
Delete mu-plugins/10up-plugin/includes/classes/.gitkeep

tobeycodes:
This directory is not empty so the .gitkeep is not needed


fabiankaegy (approved):
👍

PR #256, opened 2024-11-28
Spatie Ignition for Debugging

darylldoyle:
### Description of the Change
This PR adds a new plugin to the scaffold. The plugin adds [`spatie/ignition`](https://github.com/spatie/ignition/) as a debugging tool. Ignition is the default error page for Laravel Projects, and for good reason. It includes great stack traces as well as other helpful information that can be used to debug the issue. There's an example of an issue page below:
wp-scaffold test_

It lets engineers quickly jump to the problem code, reducing debugging time (see video below).

https://github.com/user-attachments/assets/73e82430-8837-4f82-819c-fc4a5e3cc299

We load it using the `000-wp-ignition-loader.php` file as it needs to run as soon as possible in order to catch errors in MU plugins.

Whilst this should be ignored from deployments to production environments, since we should be handling errors transparently there, it does have some protections in place to ensure it doesn't run, as long as the `wp_get_environment_type()` returns correctly.

> [!NOTE]
> This update doesn't handle Fatal Errors, like Abstract Class extension issues because they happen at compile-time and therefore are handled by PHP's and WordPress's `register_shutdown_function()` before this package has time to process them.

### How to test the Change
Clone locally. Change something in the plugins/theme to cause a PHP error. See Ignition.

### Changelog Entry
* Added - Ignition as an error page.

### Credits

Props @darylldoyle 

### Checklist:

- [ ] ~I have updated the documentation accordingly.~
- [ ] ~I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.~


mu-plugins/wp-ignition/wp-ignition.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @package WP_Ignition
+ */
+
+require __DIR__ . '/vendor/autoload.php';
+
+// Register Ignition
+\Spatie\Ignition\Ignition::make()
+	->setTheme( 'dark' )
+	->shouldDisplayException( 'production' !== wp_get_environment_type() && true === WP_DEBUG )
```
tobeycodes: **question:** is this needed if this file is never loaded because of https://github.com/10up/wp-scaffold/pull/256/files#diff-4ea83a254277eb00b13e53d487e576481ad61df4a542af71e5ed0df5df2f9333R19

mu-plugins/wp-ignition/composer.json
tobeycodes, line 1: **question:** why not include this is a dev dependency in the root composer.json so it's one less composer.json file. When running --production it won't be loaded and then the plugin code would never run

mu-plugins/000-wp-ignition-loader.php
tobeycodes, line 1: **thought:** Maybe https://github.com/roots/bedrock-autoloader ?

tobeycodes (approved):
Love this! I left my thoughts/questions in some comments but not blocking from moving forward

mu-plugins/wp-ignition/wp-ignition.php
darylldoyle: It's mostly there as a backup in case someone _does_ deploy it, or if they pull it into a pre-existing project.

mu-plugins/wp-ignition/composer.json
darylldoyle: I played with that idea a bunch. Unfortunately we'd need to load our autoloader in WP-config.php or something, otherwise it doesn't load early enough. 

Our current MU plugin loads the autoloader late in the WP boot process and we want it to load as the first plugin, to ensure all errors are handled by Ignition

mu-plugins/000-wp-ignition-loader.php
darylldoyle: Unfortunately it loads after all other MU plugins and we need this to be the first

mu-plugins/wp-ignition/composer.json
tobeycodes: I think wp-config.php is the best place to load this to be honest. The repo could have a wp-config-sample.php

mu-plugins/000-wp-ignition-loader.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Plugin Name:       WP Ignition
+ * Description:       Adds Laravel Ignition to WordPress to aid with debugging.
+ * Version:           1.0.0
+ * Author:            10up
+ * Author URI:        https://10up.com/
+ * License:           GPL v2 or later
+ * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+ * Text Domain:       wp-ignition
+ * Update URI:        https://yourwebsite.com/wp-ignition
```
claytoncollie: @darylldoyle Can we set this to `false` instead of needing to update for every client site?

mu-plugins/wp-ignition/wp-ignition.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Plugin Name:       WP Ignition
+ * Description:       Adds Laravel Ignition to WordPress to aid with debugging.
+ * Version:           1.0.0
+ * Author:            10up
+ * Author URI:        https://10up.com/
+ * License:           GPL v2 or later
+ * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+ * Text Domain:       wp-ignition
+ * Update URI:        https://yourwebsite.com/wp-ignition
```
claytoncollie: @darylldoyle Same here. Can we set to `false` to avoid needing to update for every project?

mu-plugins/wp-ignition/composer.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "10up/wp-ignition",
+  "type": "wordpress-plugin",
```
claytoncollie: @darylldoyle should this be `wordpress-muplugin` or `wordpress-mu-plugin` in case the project has specific installer paths for each type of wordpres plugin?

mu-plugins/000-wp-ignition-loader.php
darylldoyle: Good call, I didn't realise `false` was an option, but _did_ mean to update it to point to this repo. Using `false` going forward :)

mu-plugins/wp-ignition/composer.json
darylldoyle: Yup, updated to `wordpress-muplugin`

PR #257, opened 2024-11-29
Fix/return types

darylldoyle:
### Description of the Change
This PR goes through our scaffold and adds the correct return types to methods and functions. This will increase the reliability of static analysis available to engineers, either via their IDEs or a tool like PHPStan.

Additionally, it ensures that _we_, as a company, are starting engineers off on the best foot possible, with no potential errors caused by incorrect return types.

Finally, there are some places where I've added additional defensive coding to ensure that we don't inadvertently trigger errors.

### How to test the Change
Test the scaffold locally.

### Changelog Entry
* Fixed - Return types

### Credits

Props @darylldoyle 

### Checklist:

- [ ] ~I have updated the documentation accordingly.~
- [ ] ~I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.~

### DIFF
--- a/mu-plugins/10up-plugin/includes/core.php
+++ b/mu-plugins/10up-plugin/includes/core.php
@@ -96,7 +96,7 @@ function deactivate() {
 /**
  * The list of knows contexts for enqueuing scripts/styles.
  *
- * @return array
+ * @return array<string>
  */
 function get_enqueue_contexts() {
 	return [ 'admin', 'frontend', 'shared' ];
--- a/mu-plugins/10up-plugin/includes/core.php
+++ b/mu-plugins/10up-plugin/includes/core.php
@@ -108,12 +108,12 @@ function get_enqueue_contexts() {
  * @param string $script Script file name (no .js extension)
  * @param string $context Context for the script ('admin', 'frontend', or 'shared')
  *
- * @return string|WP_Error URL
+ * @return string URL
  */
 function script_url( $script, $context ) {
 
 	if ( ! in_array( $context, get_enqueue_contexts(), true ) ) {
-		return new WP_Error( 'invalid_enqueue_context', 'Invalid $context specified in TenUpPlugin script loader.' );
+		throw new \RuntimeException( 'Invalid $context specified in TenUpPlugin script loader.' );
 	}
 
 	return TENUP_PLUGIN_URL . "dist/js/{$script}.js";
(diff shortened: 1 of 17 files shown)


tobeycodes (approved):
This looks good. I'd love to see in a follow up PR in the future
- Removal of mixed
- Removal of phpstan ignore errors
- Consistent return types on get_asset_info

mu-plugins/10up-plugin/includes/core.php
```diff
@@ -108,12 +108,14 @@ function get_enqueue_contexts() {
...
- * @return string|WP_Error URL
+ * @throws \RuntimeException If an invalid $context is specified.
+ *
+ * @return string URL
  */
 function script_url( $script, $context ) {
 
 	if ( ! in_array( $context, get_enqueue_contexts(), true ) ) {
-		return new WP_Error( 'invalid_enqueue_context', 'Invalid $context specified in TenUpPlugin script loader.' );
+		throw new \RuntimeException( 'Invalid $context specified in TenUpPlugin script loader.' );
```
fabiankaegy, line 118: How do these get reported in WP? Anything to be aware of relating to this change?

mu-plugins/10up-plugin/includes/utility.php
```diff
@@ -18,17 +18,22 @@
...
 function get_asset_info( $slug, $attribute = null ) {
 	if ( file_exists( TENUP_PLUGIN_PATH . 'dist/js/' . $slug . '.asset.php' ) ) {
 		$asset = require TENUP_PLUGIN_PATH . 'dist/js/' . $slug . '.asset.php';
 	} elseif ( file_exists( TENUP_PLUGIN_PATH . 'dist/css/' . $slug . '.asset.php' ) ) {
 		$asset = require TENUP_PLUGIN_PATH . 'dist/css/' . $slug . '.asset.php';
 	} else {
-		return null;
+		$asset = [
+			'version'      => TENUP_PLUGIN_VERSION,
+			'dependencies' => [],
```
fabiankaegy, line 31: Why didn't we do this in the first place? Nice addition!

fabiankaegy (approved):
Agree with all that @tobeycodes already mentioned  👍

mu-plugins/10up-plugin/includes/core.php
darylldoyle: Unless it's caught, which this isn't, it will cause a fatal uncaught exception error. IMO, that's more desirable than a `WP_Error`, which doesn't kill the script's execution but would just cause oddities elsewhere as we try to use it. 

If we're passing an incorrect context, this should be made immediately known to the engineer, this change does that. Also, when paired with #256 we'd see a nice error page showing us where this originated from.

tlovett1:
Looks good

PR #266, opened 2025-01-06
Move Overrides out of theme and into functionality plugin

### LINKED ISSUE #249: Move non theme related code to the plugin
The README mentions:

> Functionality should be built into the 10up must-use functionality as much as possible. Presentation should be kept in the theme. Separating these two makes long term development, maintenance, and extensibility much easier.

However there are several places in the theme which have code not related to presentation. Moving this code makes it easier for projects with multiple themes to share code and share a mu-plugin rather than duplicate code and increase maintenance effort.

https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/includes/overrides.php

claytoncollie:
### Description of the Change
Small refactor to move functionality out of the theme and into the mu-plugin.  Also, after some testing, the original code was not be instantiated in the theme at all. This change not only moves it but also gets it working properly again.

Closes #249 

### How to test the Change
1. Check out `trunk`
2. View the source on the frontend
3. Search for `wp-emoji-styles-inline-css`
4. Verify it does exist
5. Checkout this branch
6. Refresh the source
7. Verify that `wp-emoji-styles-inline-css` does NOT exist

### Credits
Props @claytoncollie

### Checklist:


tobeycodes (approved):
Looks great to me. Aside from this I think we should review these actions and see if they are still relevant in 2025.

darylldoyle (approved):

Issue #247, opened 2024-11-04
Replace function files with module classes


tobeycodes:
https://github.com/10up/wp-scaffold/tree/trunk/themes/10up-theme/includes

We should be consistent with how we use actions/filters. Either we should use functions everywhere or classes everywhere.

darylldoyle:
Thanks, @tobeycodes. I've added this to the Internal Tools working group agenda to be discussed. If you have any additional context/information that would be useful for those reviewing this, I'd greatly appreciate it if you could add it.

From my perspective, the issue here is that we have classes containing actions/filters, but then we also have functional PHP files containing actions/filters. The issue with mixing them is that:

1. We end up with poorly organised code because engineers disagree _about where_ the best place to put them is.
2. Mixing functional and OOP code can be confusing when we could use OOP for everything.
3. Having one or the other would make it easier to implement tests across projects/the scaffold as we'd need less manual loading.

Is there anything there you think I've missed?

tobeycodes:
@darylldoyle Summed up perfectly

darylldoyle:
This is completed in #273, closing now

PR #273, opened 2025-01-30
Update/oop all the things

### LINKED ISSUE #251: Organize PHP classes folder
I would love to see the following subtle improvements to the classes folder

- Remove the helpers folder
- Convert get_module function into a function in the Helper namespace Helper/Module::get_module
- Remove the classes folder and move all files up one level
- Convert core.php and utility.php to classes within the namespace
- Move Module and ModuleInitialization to it's own namespace
- **suggestion:** use src instead of includes to align more with the PHP community in general

darylldoyle:
### Description of the Change
This PR moves the wp-scaffold to use a OOP approach that more closely aligns with industry standards. Currently we have classes containing actions/filters, but we also have functional PHP files containing actions/filters. The issue with mixing them is that:

* We end up with poorly organised code because engineers disagree about where the best place to put them is.
* Mixing functional and OOP code can be confusing when we could use OOP for everything.
* Having one or the other would make it easier to implement tests across projects/the scaffold as we'd need less manual loading.

Closes #251, #247 

### How to test the Change
Ensure everything still works as expected

### Changelog Entry
Developer - Moved to a more OOP structure.

### Credits
Props @darylldoyle, @tobeycodes 

### Checklist:

- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.


mu-plugins/10up-plugin/plugin.php
tobeycodes, line 1: At the top of this file we should throw an error if `\TenUpPlugin\PluginCore` does not exist telling the developer to do a `composer install`

mu-plugins/10up-plugin/plugin.php
```diff
@@ -41,13 +41,11 @@
 	require_once TENUP_PLUGIN_PATH . 'vendor/autoload.php';
 }
 
-// Include files.
-require_once TENUP_PLUGIN_INC . '/utility.php';
-require_once TENUP_PLUGIN_INC . '/core.php';
+$plugin_core = \TenUpPlugin\PluginCore::instance();
```
tobeycodes: Does this need to be a singleton?

mu-plugins/10up-plugin/src/Assets.php
```diff
@@ -0,0 +1,273 @@
...
+class Assets implements ModuleInterface {
+
+	use Module;
+
+	/**
+	 * Can this module be registered?
+	 *
+	 * @return bool
+	 */
+	public function can_register() {
```
tobeycodes, line 29: Could this be the default?

mu-plugins/10up-plugin/src/Assets.php
```diff
@@ -0,0 +1,273 @@
...
+
+	/**
+	 * Add async/defer attributes to enqueued scripts that have the specified script_execution flag.
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/12009
+	 * @param string $tag    The script tag.
+	 * @param string $handle The script handle.
+	 * @return string|null
+	 */
+	public function script_loader_tag( $tag, $handle ) {
```
tobeycodes, line 223: Wasn't this supported a while ago? https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/

mu-plugins/10up-plugin/src/Assets.php
```diff
@@ -0,0 +1,273 @@
...
+		);
+	}
+
+	/**
+	 * Enqueue editor styles. Filters the comma-delimited list of stylesheets to load in TinyMCE.
+	 *
+	 * @param string $stylesheets Comma-delimited list of stylesheets.
+	 * @return string
+	 */
+	public function mce_css( $stylesheets ) {
```
tobeycodes, line 203: Is this still needed? When do projects not use guteneberg?

mu-plugins/10up-plugin/src/Assets.php
```diff
@@ -0,0 +1,273 @@
...
+	public function can_register() {
+		return true;
+	}
+
+	/**
+	 * Register any hooks and filters.
+	 *
+	 * @return void
+	 */
+	public function register() {
```
tobeycodes, line 38: I would love to see us automate enqueueing scripts/styles. we can find all the scripts/styles in the dist folder and should be able to infer their context (or add it easily)

mu-plugins/10up-plugin/src/PluginCore.php
tobeycodes, line 1: I'd move this to the framework. For the variables such as `tenup-plugin` move these to constants the theme controls. The plugin itself can extend the class and override things when needed e.g. register

tobeycodes (approved):
Looks great!

mu-plugins/10up-plugin/src/Assets.php
claytoncollie: Please please please remove this code. I don't know a single person who uses this functionality.

mu-plugins/10up-plugin/src/Assets.php
claytoncollie: I really hope we can remove this as well.

mu-plugins/10up-plugin/src/Core/Emoji.php
```diff
@@ -1,19 +1,21 @@
...
 use TenupFramework\ModuleInterface;
 
 /**
- * Overrides class to manage WordPress core behavior modifications.
+ * Emoji
+ *
+ * @package TenUpPlugin\Core
  */
-class Overrides implements ModuleInterface {
+class Emoji implements ModuleInterface {
```
claytoncollie, line 18: Can we move this to the framework and register it by default?

mu-plugins/10up-plugin/src/Core/HeadOverrides.php
```diff
@@ -0,0 +1,51 @@
...
+
+use TenupFramework\Module;
+use TenupFramework\ModuleInterface;
+
+/**
+ * Overrides class to manage WordPress core behavior modifications.
+ *
+ * @package TenUpPlugin\Core
+ */
+class HeadOverrides implements ModuleInterface {
```
claytoncollie, line 18: Can we move this to the framework?

mu-plugins/10up-plugin/src/Assets.php
darylldoyle: Agreed, I don't think this is the right place to do that though. This is primarily around changing the structure of the scaffold, so I'm trying to move things 1:1. Let's spin up an issue to do this though!

mu-plugins/10up-plugin/src/Assets.php
darylldoyle: That change would need to be made in the framework, but I don't think it's a bad idea 🤔

mu-plugins/10up-plugin/plugin.php
darylldoyle: Good call, I've changed that!

mu-plugins/10up-plugin/src/Assets.php
darylldoyle: As above, let's spin up an issue to remove this. This PR is about the structure

mu-plugins/10up-plugin/src/Assets.php
darylldoyle: This would be an amazing enhancement. CC @fabiankaegy Toby gave me some thoughts on this in Slack too.

mu-plugins/10up-plugin/src/PluginCore.php
darylldoyle: I'm not against this as an enhancement, but I'd love to understand your thinking around overrides etc.

mu-plugins/10up-plugin/src/Core/Emoji.php
darylldoyle: Can you open that as an issue on the `wp-framework` repo, please? I think that's a great idea, but this PR isn't the right place.

mu-plugins/10up-plugin/src/Core/HeadOverrides.php
darylldoyle: As above, can you open that as an issue on the `wp-framework` repo, please? I think that's a great idea, but this PR isn't the right place.

themes/10up-block-theme/package.json
```diff
@@ -31,6 +31,9 @@
 			"editor-style-overrides": "./assets/css/editor-style-overrides.css",
 			"frontend": "./assets/js/frontend.js",
 			"block-extensions": "./assets/js/block-extensions.js"
-		}
+		},
+    "paths": {
+      "blocksDir": "./blocks"
+    }
```
fabiankaegy: Nit: It seems the indentation here is of

themes/10up-block-theme/src/Assets.php
```diff
@@ -0,0 +1,140 @@
...
+	use GetAssetInfo;
+
+	/**
+	 * Can this module be registered?
+	 *
+	 * @return bool
+	 */
+	public function can_register() {
+		return true;
+	}
```
fabiankaegy, line 31: Stupid question: 

Would it not be better to have this return true by default in the `ModuleInterface` base class? In 90% of cases that is what we want. And when we don't we can override it. But it seems redundant to always have to add this to every single file 🤔 (very loosely held opinion)

themes/10up-block-theme/src/Assets.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Register any hooks and filters.
+	 *
+	 * @return void
+	 */
+	public function register() {
+		add_action( 'init', [ $this, 'scripts' ] );
+		add_action( 'init', [ $this, 'register_all_icons' ], 10 );
+		add_action( 'wp_enqueue_scripts', [ $this, 'styles' ] );
+		add_action( 'enqueue_block_editor_assets', [ $this, 'editor_style_overrides' ] );
```
fabiankaegy, line 42: Question:

Does this actually have the proper IDE integration? Based on what @tobeycodes said before this here is as bad as the `$n()` function we used to have 🤔

themes/10up-block-theme/src/Assets.php
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	/**
+	 * Enqueue scripts for front-end.
+	 *
+	 * @return void
+	 */
+	public function scripts() {
+
```
fabiankaegy: nit: to stay consistent lets remove this empty line

themes/10up-block-theme/src/Blocks.php
```diff
@@ -0,0 +1,124 @@
...
+use TenupBlockTheme\Traits\GetAssetInfo;
+
+/**
+ * Blocks module.
+ *
+ * @package TenupBlockTheme
+ */
+class Blocks implements ModuleInterface {
+
+	use Module;
```
fabiankaegy, line 21: Stupid question: 

Why do we need to add `use Module` here? Isn't that already coming from the `implements ModuleInterface`?

themes/10up-theme/package.json
```diff
@@ -30,7 +30,10 @@
       "editor-style-overrides": "./assets/js/admin/editor-style-overrides.js",
       "frontend": "./assets/js/frontend/frontend.js",
       "shared": "./assets/js/shared/shared.js",
-      "block-editor-script": "./includes/block-editor-script.js"
+      "block-editor-script": "./assets/js/block-editor/block-editor-script.js"
+    },
+    "paths": {
+      "blocksDir": "./blocks"
```
fabiankaegy, line 36: nit: Indentation here is of

themes/10up-theme/src/Assets.php
```diff
@@ -0,0 +1,238 @@
...
+		/*
+		 * Uncoment this to use the shared.js file.
+			wp_enqueue_script(
+				'shared',
+				TENUP_THEME_TEMPLATE_URL . '/dist/js/shared.js',
+				$this->get_asset_info( 'shared', 'dependencies' ),
+				$this->get_asset_info( 'shared', 'version' ),
+				true
+			);
+		*/
```
fabiankaegy, line 83: Doesn't need to be in this PR but I'm in favor of removing all the "shared" stuff here also...

themes/10up-theme/src/Assets.php
```diff
@@ -0,0 +1,238 @@
...
+		);
+	}
+
+	/**
+	 * Enqueue styles for admin
+	 *
+	 * @return void
+	 */
+	public function admin_styles() {
+
```
fabiankaegy: nit: remove empty line

themes/10up-theme/src/Blocks.php
```diff
@@ -0,0 +1,183 @@
...
+	 *
+	 * @return void
+	 */
+	public function register() {
+		add_action( 'enqueue_block_editor_assets', [ $this, 'blocks_editor_styles' ] );
+		add_action( 'init', [ $this, 'enqueue_block_specific_styles' ] );
+		add_action( 'init', [ $this, 'register_theme_blocks' ] );
+		add_action( 'init', [ $this, 'register_block_pattern_categories' ] );
+	}
+
```
fabiankaegy, line 44: nit: remove additional empty line

fabiankaegy (approved):
Left a few small questions / comments but overall looking forward to the new stricture :) 

Also lets make sure we add the one place for utilities as discussed in the internal projects call 👍

themes/10up-block-theme/src/Blocks.php
darylldoyle: Not a stupid question! The `ModuleInterface` tells the class what methods it _needs_ to implement, the `Module` trait does the actual implementation.

The Interface/Trait approach do the same thing as the old Module abstract class used to, without the downside of _everything_ having to inherit a single class. This makes it more flexible for engineers to change those methods when they have things they actually want to do.

themes/10up-block-theme/src/Assets.php
darylldoyle: I agree, though that's a change we'll need to make in the Framework and then update here 🙂

themes/10up-block-theme/src/Assets.php
darylldoyle: It definitely _should_ work with all decent IDE',s as this is the standard way of calling class methods in PHP. The `$n()` function was an odd one as it was a wrapper around `__NAMESPACE__` which IDE's also don't seem to like. 

This is what I see in my IDE, which is what I'd expect:

image

fabiankaegy: Just tested and in VSCode it works also. So never mind my comment here =D

PR #281, opened 2025-02-27
Update/asset loading

darylldoyle:
### Description of the Change
This PR moves the asset loading trait out of each theme and plugin and into the WP Framework. The only _real_ change, is that we need to pass some vars in to set up the paths, so that the framework knows where to asset files.

This is reliant on https://github.com/10up/wp-framework/pull/5 being merged and released.

### How to test the Change
Ensure JS and CSS files are loading correctly.

### Changelog Entry

* Developer - Changed where the GetAssetInfo trait is loading from.

### Credits

Props @darylldoyle 

### Checklist:


darylldoyle:
@fabiankaegy, the tests are failing for now, as we've not released the update in the framework, but what do you think of this approach?

fabiankaegy (commented):
Works for me :)

PR #284, opened 2025-03-24
Fix: blocks directory path in scaffold block command

fabiankaegy:
Follow up to the recent changes of moving the blocks dir top level


darylldoyle (approved):

darylldoyle:
Good catch! I missed this

PR #298, opened 2025-07-17
Move 10up-toolkit to production dependencies; Update to latest version.

### LINKED ISSUE #138: Include 10up-toolkit in NPM dependencies instead of devDependencies

claytoncollie:
### Description of the Change
This PR closes #138 by moving the 10up-toolkit from `devDependencies` to `dependencies`, making it available in the Node production environment.

I also cleaned up some files in the block theme that don't need to be there since the root config files should take precedence.

Lastly, I updated 10up-toolkit to the latest version across the entire project.

Closes #138 

### How to test the Change
Make sure you can build the workspace assets in development and production environments. Run the following commands in your terminal in the wp-content directory.

1. `export NODE_ENV=development`
2. `npm install`
3. `npm run build`
4. Verify assets are built properly.

1. `export NODE_ENV=production`
2. `npm install`
3. `npm run build`
4. Verify assets are built properly.

### Credits
Props @claytoncollie @johnwatkins0 @Antonio-Laguna 

### Checklist:


fabiankaegy (approved):
This makes sense :) Thanks for the fix!

PR #301, opened 2025-07-18
Add scrollbar and JS detection to Classic Theme

claytoncollie:
Small change in the block theme to match the classic theme in placement and content.

### Credits
Props @claytoncollie

### Checklist:

### DIFF
--- a/themes/10up-block-theme/src/ThemeCore.php
+++ b/themes/10up-block-theme/src/ThemeCore.php
@@ -86,17 +86,7 @@ class ThemeCore {
 		do_action( 'tenup_block_theme_init' );
 	}
 
-	/**
-	 * Handles JavaScript detection.
-	 *
-	 * Adds a `js` class to the root `<html>` element when JavaScript is detected.
-	 *
-	 * @return void
-	 */
-	public function js_detection() {
-
-		echo "<script>(function(html){html.className = html.className.replace(/\bno-js\b/,'js')})(document.documentElement);</script>\n";
-	}
+	
 
 	/**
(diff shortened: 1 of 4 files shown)


themes/10up-block-theme/src/ThemeCore.php
```diff
@@ -86,32 +83,6 @@ function () {
...
-	 *
-	 * Adds a JavaScript event listener to the DOMContentLoaded event. When the DOM is fully loaded,
-	 * it calculates the width of the scrollbar and sets a CSS variable `--wp--custom--scrollbar-width` with the width.
-	 * It also adds an event listener to the window resize event to update the scrollbar width when the window is
-	 * resized.
-	 *
-	 * @return void
-	 */
-	public function scrollbar_detection() {
-		echo '<script>window.addEventListener("DOMContentLoaded",()=>{const t=()=>window.innerWidth-document.body.clientWidth;const e=()=>{document.documentElement.style.setProperty("--wp--custom--scrollbar-width",`${t()}px`)};e();});</script>' . "\n";
```
fabiankaegy: Id be in favor of keeping this. If we want to remove it we need to also remove its counter parts from theme.json. 

But I think it has value to be there

claytoncollie: Ah, yes. Good point. I saw it was removed from the scaffold, so I thought it was good to remove here. Maybe this PR needs to change to add it back to the classic theme.

claytoncollie:
@fabiankaegy The classic theme now matches the block theme

fabiankaegy (approved):

PR #313, opened 2026-02-09
Feature: Add Scaffold Command

### LINKED ISSUE #204: Automate the process of renaming the theme and plugin to match the project name

When starting a new project, engineers should change the name of the theme and plugin to match the project's name. This includes changing names in many places, like the theme folder, the domain, the theme name in the style.css file, and other parts.

It's best to have a clear, standard way to do this, like using a script or following a step-by-step guide.

### Designs

_No response_

_No response_

### Code of Conduct

fabiankaegy:
## Add interactive project scaffold CLI

Adds a Node.js CLI script (`bin/scaffold.mjs`) that automates the initial project setup when using this repo as a starter. Instead of manually finding and replacing placeholder strings across dozens of files, developers can run `npm run scaffold` and answer a few prompts.

### What it does

The script walks the user through:

1. **Hosting platform** - Standard WordPress or WordPress VIP (determines whether the plugin lives in `mu-plugins/` or `client-mu-plugins/`)
2. **Theme type** - Block Theme (recommended) or Classic Theme. The unused theme is deleted.
3. **Project name** - A human-readable name (e.g. "Acme Corp") from which all naming variants are derived automatically.
4. **Accept or customize** - The user can accept the auto-derived values or override each one individually (namespace, constants, text domain, slug, etc.).
5. **Project metadata** - Author name, email, URI, description, composer vendor, homepage, and repository URL. The repo URL is pre-populated from the git remote when available.

Once confirmed, the script:

- Deletes the unused theme directory
- Moves the plugin to `client-mu-plugins/` when VIP is selected
- Replaces all scaffold placeholder strings across PHP, JSON, YAML, XML, Markdown, and config files
- Handles both `TenupBlockTheme` and `TenUpBlockTheme` capitalization variants
- Handles the block theme's internal use of `tenup-theme` (style handles, text domains, pattern slugs) alongside `tenup-block-theme`
- Removes the deleted theme's constants from `phpstan/constants.php` and stale paths from `phpstan.neon`
- Renames directories (`10up-plugin` to `{slug}-plugin`, `10up-block-theme` to `{slug}-theme`)
- Renames the loader file and `.pot` translation files
- Fixes the plugin `package.json` `globalStylesDir`/`globalMixinsDir` paths
- Deletes `package-lock.json` and `composer.lock` files (must be regenerated after path changes)
- Cleans up after itself by removing the `scaffold` script and `@inquirer/prompts` dependency from `package.json`

### Files changed

- **`bin/scaffold.mjs`** (new) - The scaffold CLI script
- **`package.json`** - Added `scaffold` script and `@inquirer/prompts` dev dependency

### How to test

1. Clone the repo fresh
2. Run `npm install`
3. Run `npm run scaffold`
4. Walk through the prompts (try both block theme and classic theme, standard and VIP)
5. Verify all placeholder strings have been replaced: `grep -r "tenup\|TenUp\|Tenup\|TENUP\|10up-plugin\|10up-theme\|10up-block-theme" --include="*.php" --include="*.json" --include="*.css" --include="*.neon" --include="*.xml" --include="*.yml" .`
6. Verify `composer install` and `npm install` succeed after scaffolding
7. Verify `npm run build` completes without errors

### DIFF
--- a/bin/scaffold.mjs
+++ b/bin/scaffold.mjs
@@ -21,6 +21,7 @@ import {
 	mkdirSync,
 } from 'node:fs';
 import { join, resolve, extname } from 'node:path';
+import { parseArgs } from 'node:util';
 
 // ---------------------------------------------------------------------------
 // Constants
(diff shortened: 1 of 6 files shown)


claytoncollie:
Attaching #204 to this ticket.

claytoncollie:
@fabiankaegy I really like this idea, but I am concerned about it living in this repo. Could this be its own NPM package that when finished prompts you to remove the package? Or maybe it lives in WP framework? Again, we need this functionality, but I hate inheriting a project and trying to understand if some files are needed or not to clean things up over time.

darylldoyle:
I really like the idea here. One question though: isn’t this essentially overlapping with what we already do in https://github.com/10up/10up-toolkit/tree/develop/packages/toolkit/scripts/project?


That approach works well for initial project setup, but I wonder if this is an opportunity to push the concept further.

What I have in mind is something more general and extensible. For example, a single CLI that can:

* Scaffold a new project
	`fueled new project MyProjectName --vip`
* Add a plugin or theme within an existing project
	`fueled new plugin MyNewPlugin --client-mu-plugin`
* Install shared tooling like WP Boost into an existing project


Over time, this could expand to more granular scaffolding as well, such as:
```shell
fueled new endpoint SomeRestName

fueled new taxonomy ContentType
```

The goal would be to reduce the number of bespoke commands and shell scripts engineers need to remember, and instead provide a consistent, discoverable interface.

If that direction makes sense, this might be better suited to a separate repo that can be pulled into this one for convenience, rather than living directly inside wp-scaffold.

Curious to hear your thoughts. Does this align with where you see this heading?

claytoncollie:
@darylldoyle I remember you talking about this CLI awhile back comparing it to what Laravel does. I also like that idae as it can encompass tools that backend, frontend, devops, and webapps do into a single  inferface. I wrote one of these for my company using https://oclif.io/ highly recommend that project as a starting point.

fabiankaegy:
@darylldoyle @claytoncollie thanks for the quick feedback here :) Very much appreciate you both :) 

I'm really torn myself and am not sure which path is the right one. On one hand, I don't think the scaffold commands in toolkit ever really worked well / got integrated into peoples workflows. And it'd been broken for a while with no-one supporting it... So I've been considering removing it from toolkit again to keep that core lighter... 

The reason why I thought it would make sense to have this as part of the scaffold was just colocation so you don't have to remember different tools / different places to maintain things. You can just make changes in one spot and things never get out of sync. 

I have considered adding a self delete option that is on by default which removes all of the scaffold code after you run it once 😂 

---

Regarding the possibilities of having a more fully fleshed out scaffolding tool we can use for classes, plugins etc is also something I know we've talked about lot about in the past... I just don't know if we still need that in the age of AI... Or whether the 10up/wp-framework skill is just enough to do all the work once you are past initial scaffolding... 


---

Those basically are all the points for why I thought and still kind of think that the scaffold is the best place for this to live. It is an opinion that I am happy to change if you two feel strongly that we should move it elsewhere :)

claytoncollie:
I am starting to lean more towards Fabian's approach especially with AI being abel to discovery these commands easily and AI being able to update the command if the scaffold changes. Simple solutions are starting to prove more beneficial with AI.

darylldoyle:
Thanks both, this is helpful context. I think there are really two separate threads here.



## **Toolkit vs Scaffold**

I agree with the assessment around the Toolkit script. It was never well-advertised internally, so it didn’t see real adoption, and I’m totally fine with removing it from there.

My hesitation with shipping this directly in `wp-scaffold` is more about longevity. Once it ships, it effectively becomes stale. If someone later wants to spin up a new plugin, they still have to:

* Go back to the scaffold repo
* Pull the latest version
* Manually extract the plugin/theme
* Rename and refactor things
* Drop it into an existing project

That friction adds up, and it doesn’t really solve the “ongoing scaffolding” problem, only the initial project bootstrap.



## **Scaffolding vs AI**

I get the argument that AI can generate a lot of this now, and I agree it’s powerful. My concern is about *where* we lean on AI versus where we can be more deterministic.

For things we already have strong, opinionated examples for (taxonomies, REST endpoints, post types, etc), I think there’s real value in having concrete scaffold commands rather than asking an LLM to invent structure every time. A CLI that generates known-good files gives us:

* Consistency across projects
* Fewer weird variations between models and prompts
* Less token usage for things that don’t need “thinking”

This is roughly how Laravel Boost and Filament approach it. The LLM is instructed to run CLI commands to generate scaffolds, and then it *fills them out*. That means AI is used where it shines (business logic, edge cases), while structure and conventions stay grounded in the framework we’ve already built.

So I’m not anti-AI here, just trying to be deliberate about not using it for things we already know how to do well.



## **Where this leaves me**

I still think there’s a strong case for a more general, extensible CLI that can:

* Scaffold new projects
* Add plugins/themes to existing projects
* Generate common framework primitives

That feels like something that could live independently, but be pulled into `wp-scaffold` for convenience. It avoids staleness, reduces cognitive load for engineers, and plays nicely with AI rather than competing with it.



Hopefully, that clarifies where I’m coming from, and I’m very open to refining this direction together.

fabiankaegy:
@darylldoyle That's really useful :) 

So I think that means regardless of whether we move it into the scaffold or leave it outside of the scaffold, we will run into the problem of the two things possibly diverging in both options.

If it's in the scaffold (but references upstream code from the repo for new scaffolded things) the two can become out of sync. And new enhancements to the CLI are hard to get back into the project...

If it lives outside we don't have that problem but the same scenario as currently can happen where one gets updated but the other gets forgotten. And it's harder to make the team aware of it... 


---

I'm not sure which is better / where it leaves us  😂 I can see both being good and both being bad options... 

---

In order to make things easier, I think the best option is to have the actual API interface of calling the CLI (regardless of where it lives) proxied through an NPM command in the repo root. 

So you can simply run `scaffold` / `scaffold:feature` or whatever which makes it easily discoverable 🤔 

If we go with that I think we can merge this here now and then iterate on it as we see fit. If we want to move it out we just delete the script and move proxy any external CLI through the existing API interface 🤔

bin/scaffold.mjs
```diff
@@ -0,0 +1,872 @@
...
+	'.mp3',
+	'.mov',
+	'.avi',
+	'.pdf',
+	'.doc',
+	'.docx',
+	'.lock',
+]);
+
+const SKIP_DIRS = new Set(['node_modules', 'vendor', '.git', 'plugins']);
```
claytoncollie, line 59: @fabiankaegy please skip `dist`

claytoncollie (approved):

PR #322, opened 2026-03-25
Add scaffold:vip command for VIP repo overlay

fabiankaegy:
## Summary
- Adds `npm run scaffold:vip -- --target <path>` command that overlays the 10up scaffold onto an existing WordPress VIP repo
- Intelligently merges configs (composer.json, .gitignore, .editorconfig, phpcs.xml, .deployignore) while preserving VIP-specific files (vip-config, .devcontainer, .wpvip, etc.)
- Extracts `buildReplacementMap()` into `scaffold-helpers.mjs` so replacement logic is shared between `scaffold.mjs` and the new `scaffold-vip.mjs`

### What the command does
1. Copies 10up theme, plugin, tooling configs, CI workflows, and scaffold scripts to the target VIP repo
2. Preserves VIP-specific files (vip-config/, .deployignore, .devcontainer/, .wpvip/, images/, private/, languages/)
3. Merges composer.json (combines both dev deps, removes VIP-provided debug plugins, uses GPL license)
4. Merges phpcs.xml with both WordPress-VIP-Go and 10up-Default rulesets
5. Integrates plugin loader into VIP's plugin-loader.php
6. Runs string replacements and directory renames for the project name
7. Supports all existing flags: --fueled, --ignite, --self-destruct, theme/plugin overrides

## Test plan
- [ ] Copy a fresh VIP skeleton to a temp dir and run `npm run scaffold:vip -- --target /tmp/test -n "Test Project" -t block -y`
- [ ] Verify VIP files preserved: vip-config/, .deployignore, .devcontainer/, .wpvip/
- [ ] Verify 10up files added: theme, plugin in client-mu-plugins/, .eslintrc, phpstan, CI workflows
- [ ] Verify merged configs: composer.json has both dep sets, phpcs.xml has both rulesets, .gitignore tracks package.json
- [ ] Verify string replacements: no remaining `10up-plugin`/`TenUpPlugin` strings outside scaffold scripts
- [ ] Test with `--theme classic` for classic theme variant
- [ ] Verify existing `npm run scaffold` still works (refactored to use shared `buildReplacementMap`)

### DIFF
--- a/bin/scaffold-helpers.mjs
+++ b/bin/scaffold-helpers.mjs
@@ -108,12 +108,12 @@ export function getGitOrgFromUrl(url) {
 // File-walking helpers
 // ---------------------------------------------------------------------------
 
-export function walkFiles(dir, results = []) {
+export function walkFiles(dir, results = [], skipDirs = SKIP_DIRS) {
 	for (const entry of readdirSync(dir, { withFileTypes: true })) {
 		const fullPath = join(dir, entry.name);
 		if (entry.isDirectory()) {
-			if (SKIP_DIRS.has(entry.name)) continue;
-			walkFiles(fullPath, results);
+			if (skipDirs.has(entry.name)) continue;
+			walkFiles(fullPath, results, skipDirs);
 		} else if (entry.isFile()) {
 			if (BINARY_EXTENSIONS.has(extname(entry.name).toLowerCase())) continue;
 			if (entry.name === 'package-lock.json') continue;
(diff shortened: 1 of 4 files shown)


bin/scaffold-helpers.mjs
```diff
@@ -122,3 +122,130 @@ export function walkFiles(dir, results = []) {
...
+ * Build the replacement map for project scaffold.
+ *
+ * @param {object}  options
+ * @param {boolean} options.isBlock - Whether the theme is a block theme
+ * @param {boolean} options.isVip   - Whether the hosting is VIP
+ * @param {object}  options.values  - The derived values object
+ * @param {string}  options.slug    - The project slug (kebab-case)
+ * @returns {Array<[string, string]>} Sorted replacement pairs (longest first)
+ */
+export function buildReplacementMap({ isBlock, isVip, values, slug }) {
```
claytoncollie, line 140: @fabiankaegy Is this starting to be too many parameters? maybe we switch to an array so it can grow over time.

bin/scaffold-helpers.mjs
```diff
@@ -122,3 +122,130 @@ export function walkFiles(dir, results = []) {
 	}
 	return results;
 }
+
+// ---------------------------------------------------------------------------
+// Replacement map builder (shared between scaffold.mjs and scaffold-vip.mjs)
+// ---------------------------------------------------------------------------
+
+/**
+ * Build the replacement map for project scaffold.
```
claytoncollie, line 131: @fabiankaegy I already created an issue for it but looking through this list emphasis that we need to go through every string and condense and simplify. Some of these replacements are a simple captalization difference. Super annoying to have to find an account for all of those.

bin/scaffold-vip.mjs
```diff
@@ -0,0 +1,1084 @@
...
+    --author-uri <uri>           Author URI
+    --description <text>         Project description
+    --composer-vendor <vendor>   Composer vendor slug
+    --homepage-url <url>         Project homepage URL
+    --repo-url <url>             Repository URL
+
+  Ignite:
+    --fueled                     Mark as a Fueled/10up project (defaults to block theme + Ignite)
+    --ignite                     Install Ignite WP plugins after scaffolding
+    --skip-ignite                Skip Ignite installation (useful with --fueled)
```
claytoncollie, line 137: @fabiankaegy This feels confusing. Not sure of the solution. I understand we want to push ignite hard but maybe --fueled and --ignite need to be there own commands so you dont have to do something like --fueled --skip-ignite . Maybe I also blowing this out of propoertion and 99% of the time we use ignite and it doesnt matter.

package.json
```diff
@@ -26,7 +26,8 @@
     "test": "npm run test --workspaces --if-present",
     "clean-dist": "npm run clean-dist --workspaces --if-present",
     "scaffold": "node bin/scaffold.mjs",
-    "scaffold:plugin": "node bin/scaffold-plugin.mjs"
+    "scaffold:plugin": "node bin/scaffold-plugin.mjs",
+    "scaffold:vip": "node bin/scaffold-vip.mjs"
```
claytoncollie, line 30: @fabiankaegy Would be interesting to see if other hosting providers need this sort of special treatment with repo structure and also match up to the ci-library on how they get deployed. I know we have a few specific hosts called out in that project.

pdavies88:
@fabiankaegy just tried this on an init commit of a recent VIP repo and worked perfect: https://github.com/wpcomvip/dhcs-ca-gov/commit/8bccf5f935420854f79fff6d04ba370e87df6aaa

package.json
pdavies88: The other two which I see a lot are WP Engine and Pantheon not sure if those give starter builds and then pass them off without running scaffold first. I think we could do these if we needed in a separate PR to keep this one focused on VIP

bin/scaffold-vip.mjs
pdavies88: Pretty sure this is for the CLI options so that we can skip if we want example:
```
> node bin/scaffold-vip.mjs --target /Users/preston10up/Local Sites/dhcs2/app/public/wp-content/


  WordPress VIP Overlay Scaffold


  Warning: Target repo has uncommitted changes.
  Consider committing first so the overlay can be reviewed as a single diff.

✔ Which theme type would you like to use? Classic Theme
✔ Is this a Fueled / 10up project? Yes

  Note: Block themes are recommended for Fueled/10up projects.

✔ Install Ignite WP plugins? (Recommended for Fueled/10up projects) No
✔ Project name (human-readable, e.g. "Acme Corp"): DHCS
```

pdavies88 (approved):

PR #329, opened 2026-05-05
fix: output dir="ltr" attribute for LTR languages

### LINKED ISSUE #174: Fix support for CSS logical properties

Currently, when using CSS logical properties, postcss creates two sets of rules, one for `html[dir="ltr"]` and one for `html[dir="rtl"]`. This would be fine, however, the native WordPress [language_attributes()](https://developer.wordpress.org/reference/functions/language_attributes/) function only adds a `dir` attribute to `html` if the language direction is RTL.

I would like to see `overrides.php` in the theme include a filter on `language_attributes` to add `dir="ltr"` to the language attributes when `is_rtl()` is false.

### Designs

An example of the function that adds this language attribute:

```php
/**
 * Add the dir attribute to the body tag if it's LTR. WordPress doesn't do this by default, and postcss scopes CSS logical properties to the text direction.
 * This means that if the text direction is LTR, the CSS logical properties will not be applied.
 *
 * @see https://developer.wordpress.org/reference/functions/get_language_attributes/
 * @param string $output attribues to add to the body tag.
 *
 * @return string Updated attributes.
 */
function maybe_add_ltr_language_attribute( $output ) {
	// If the current language is LTR, add the dir attribute.
	if ( ! is_rtl() ) {
		$output .= ' dir="ltr"';
	}

	return $output;
}
```

An alternative could be modifying the default postcss config to handle logical properties in a way where RTL styles are scoped to `html[dir="rtl"]`, while LTR styles are unscoped.

### Code of Conduct

fabiankaegy:
## Summary
- Adds a `language_attributes` filter to both `10up-theme` and `10up-block-theme` that appends `dir="ltr"` when `is_rtl()` is false.
- WordPress core only emits a `dir` attribute on `<html>` for RTL sites, but `postcss-logical` scopes generated rules to both `html[dir="ltr"]` and `html[dir="rtl"]`. Without this, LTR-scoped rules never match.

Fixes #174

## Test plan
- [ ] Load a page on a default LTR site and confirm `<html>` includes `dir="ltr"`.
- [ ] Switch the site to an RTL locale (e.g. `ar`) and confirm `<html>` includes `dir="rtl"` (and not both).
- [ ] Verify CSS rules generated by `postcss-logical` under `html[dir="ltr"]` now apply on LTR sites.
- [ ] Repeat the above for both `10up-theme` and `10up-block-theme`.

### DIFF
--- a/themes/10up-block-theme/src/ThemeCore.php
+++ b/themes/10up-block-theme/src/ThemeCore.php
@@ -29,6 +29,8 @@ class ThemeCore {
 		add_action( 'wp_head', [ $this, 'js_detection' ], 0 );
 		add_action( 'wp_head', [ $this, 'scrollbar_detection' ], 0 );
 
+		add_filter( 'language_attributes', [ $this, 'add_ltr_language_attribute' ] );
+
 		do_action( 'tenup_block_theme_loaded' );
 	}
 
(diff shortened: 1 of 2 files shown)


claytoncollie (approved):
I personally don't like the js_detection, scrollbar_detection or this language_attribute, BUT I know they are valuable and you have a good reason to include them. APPROVED!

